PR #5, opened 2020-07-17
Update sitebase.jinja2

specsdom:
### Master List of Changes ###

**sitebase.jinja2 edits**
- Softening language of notice that site is still changing (participants have reported confusion that they cannot do anything on the website because it is "undergoing revisions")
- Moved positions of text and clickable elements
- Creating a nav bar that nests the breadcrumbs elements and moves "Log Out" to the right side; this is in part to ensure users don't accidentally click on this link, as it's presented as the one of the first items in every page's main content

**signup_intermediate.jinja2 edits**
- Made changes to content, formatting, and moved positions of text, as per Kat's suggestions

**home.jinja2**
- Made changes to content, formatting, and moved positions of text
- Switched out "Refresh" button to be view AuthRocket profile.  I don't see any use case of a "Refresh" button as the pathway for participants would likely be: They see this page -> Check their email -> Find confirmation email -> Confirm email -> They go back to this page, which has already been refreshed for different content

### DIFF
--- a/microsetta_interface/templates/home.jinja2
+++ b/microsetta_interface/templates/home.jinja2
@@ -24,23 +24,23 @@
         {% endif %}
     </div>
     {% else %}
-    <p>
-        It appears that your email has not yet been verified.
-        Please check your email account (and spam folder) for a verification email
-        from "The Microsetta Initiative" and follow its instructions.
-        This email will come from our authentication service AuthRocket,
-        and use the email address "noreply@loginrocket.com".
+  <p>
+        To continue: Check your email account to obtain your confirmation email and link, which was sent to confirm your account.
+        This email will come from our authentication service, AuthRocket, under the email address "noreply@loginrocket.com".
     </p>
     <p>
-        If you cannot locate the original verification email, have it resent by
-        clicking <a href = "{{ authrocket_url }}/profile">here</a> to view your AuthRocket profile and then
-        clicking the "resend verification email" link.
+        If you cannot locate the original confirmation email, have it resent by
+        clicking the button below to view your AuthRocket profile, and then
+        clicking the "resend verification email" link next to your email address.
     </p>
     <p>
-        <button class = "btn btn-primary" onclick="window.location.replace('{{authrocket_url}}/login?redirect_uri={{endpoint}}/authrocket_callback');">
-            Refresh
-        </button>
+        <a href = "{{ authrocket_url }}/profile"><button class = "btn btn-primary">
+            View AuthRocket Profile
+        </button></a>
     </p>
+    
+    <p>If you do not receive this email within a few minutes and have checked your spam/junk mail folders, please contact us at
+    <a href="mailto:microsetta@ucsd.edu">microsetta@ucsd.edu</a>.
     {% endif %}
 {% else %}
     You are not logged in. Please click to either: <br/>
--- a/microsetta_interface/templates/signup_intermediate.jinja2
+++ b/microsetta_interface/templates/signup_intermediate.jinja2
@@ -20,8 +20,9 @@
     <img src="/static/img/old_registration_no_password.jpg" class="resize"/><br/>
     Some cards may also contain a 'Password:' field, which we currently DO NOT require, so please disregard it.<br/>
     <img src="/static/img/old_registration_password.jpg" class="resize"/>
+    </p>
     
        <p>Continue to Sign Up: <br/>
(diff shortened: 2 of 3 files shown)


microsetta_interface/templates/sitebase.jinja2
```diff
@@ -85,19 +84,19 @@
...
         <br />
         <hr>
+          <p>We will periodically update this site with minor revisions; we
+          apologize if you encounter any temporary inconveniences.</p>
         {% if not disable_faq %}
         <p>
-            Questions?  Check our <a href="/view_faq" target="_blank">FAQ</a>!
-        </p>
+            Questions?  Check out our <a href="/view_faq" target="_blank">FAQ</a>!</br>
         {% endif %}
```
dhakim87, line 92: Since the {% if ... %} {% endif %} surrounds the <p> tag, this looks wrong to me.  It means that if I render this page with disable_faq set to True, the <p> tag will not be rendered (but the </p> on line 98 will be).  We should always have matched tags within the {% %} control flow constructs.  

You can fix this in this case by moving the <p> up above the {% if not disable_faq %}

microsetta_interface/templates/sitebase.jinja2
```diff
@@ -55,6 +54,9 @@
     <div class="content">
 
     {% if show_breadcrumbs %}
+        <nav class="navbar navbar-expand-lg navbar-light" style="background-color:#e9ecef;">
```
wasade, line 57: These changes with the navbar won't work with the current conditional logic of the template unfortunately. Specifically, the closing `</nav>` on line 70 is not assured to pair with the opening one on line 57. The rendering also looks a little less clean, although I'm not sure if that is due to nested `nav` tags. 

Would it be possible to remove these changes, and we consider them for a separate PR?

microsetta_interface/templates/home.jinja2
```diff
@@ -24,23 +24,23 @@
...
-        clicking the "resend verification email" link.
+        If you cannot locate the original confirmation email, have it resent by
+        clicking the button below to view your AuthRocket profile, and then
+        clicking the "resend verification email" link next to your email address.
     </p>
     <p>
-        <button class = "btn btn-primary" onclick="window.location.replace('{{authrocket_url}}/login?redirect_uri={{endpoint}}/authrocket_callback');">
-            Refresh
-        </button>
+        <a href = "{{ authrocket_url }}/profile"><button class = "btn btn-primary">
```
wasade, line 37: I'm really concerned about whether this introduces a functional change as it removes the `onclick` behavior. What was the intention with this change?

microsetta_interface/templates/signup_intermediate.jinja2
```diff
@@ -3,20 +3,25 @@
...
+       verify we have the correct information. If you do not receive this email within a few minutes and have checked your spam/junk mail folders, please contact us at <a href="mailto:microsetta@ucsd.edu">microsetta@ucsd.edu</a>, 
+       and we will resolve the issue.<p>
 </div>
-<h2>Registration card</h2>
+<h4><u>Registration Card</u></h4>
 <div class="list-group">
-    Participant kits contain a registration card that looks like the following. You do not need this card to create a login, however you will need this card afterward in order to create your account.
-    <img src="/static/img/old_registration_no_password.jpg" class="resize"/>
-    If your registration card contains a "Password:" field as shown in the following image, please disregard it.
+    <p>The kit you have received contains a registration card (found inside the kit's plastic sleeve for Step #2). On this card, you will find
```
wasade, line 18: I believe this is missing a closing `</p>`

microsetta_interface/templates/signup_intermediate.jinja2
```diff
@@ -3,20 +3,25 @@
...
+    your Kit ID, which you will need to create your account and to later connect your samples to your profile.<br/>
+    <img src="/static/img/old_registration_no_password.jpg" class="resize"/><br/>
+    Some cards may also contain a 'Password:' field, which we currently DO NOT require, so please disregard it.<br/>
     <img src="/static/img/old_registration_password.jpg" class="resize"/>
+    
+       <p>Continue to Sign Up: <br/>
+       <a href="{{authrocket_url}}/signup?redirect_uri={{endpoint}}/authrocket_callback"><button class="btn btn-primary">Sign Up</button></a>. </p>
```
wasade, line 25: The "." at the end places a period after the button, is that intended?

wasade:
Thanks, @specsdom!!

PR #11, opened 2020-07-23
Integrated public api for a new sample results page

dhakim87:
Added results page from source template.  

Screen Shot 2020-07-23 at 3 38 31 PM


microsetta_interface/implementation.py
```diff
@@ -949,6 +951,30 @@ def post_update_sample(*, account_id=None, source_id=None, sample_id=None):
...
 
 
+@prerequisite([SOURCE_PREREQS_MET])
+def get_sample_results(*, account_id=None, source_id=None, sample_id=None):
+    has_error, source_output, _ = ApiRequest.get(
+        '/accounts/%s/sources/%s' %
+        (account_id, source_id)
+    )
+    if has_error:
+        return source_output
```
wasade, line 961: This was previously unsafe, see #10

microsetta_interface/implementation.py
```diff
@@ -949,6 +951,30 @@ def post_update_sample(*, account_id=None, source_id=None, sample_id=None):
...
+        (account_id, source_id, sample_id))
+    if has_error:
+        return sample_output
+
+    return _render_with_defaults('sample_results.jinja2',
+                                 account_id=account_id,
+                                 source_id=source_id,
+                                 sample=sample_output,
+                                 source_name=source_output['source_name'],
+                                 taxonomy=SERVER_CONFIG["taxonomy_resource"],
```
wasade, line 974: I'm not sure I understand why this is a server configuration detail. What happens when shotgun data are available?

microsetta_interface/server_config.json
```diff
@@ -1,6 +1,9 @@
 {
   "endpoint": "http://localhost:8083",
   "private_api_endpoint": "http://localhost:8082/api",
+  "public_api_endpoint": "http://localhost:8084/results-api",
+  "taxonomy_resource": "greengenes",
```
wasade, line 5: Aren't the availalbe resources discoverable from the public api?

microsetta_interface/static/css/4_column_flex.css
```diff
@@ -0,0 +1,29 @@
+  * {
```
wasade, line 1: It looks like the contents of this file are copied from somewhere else on the web. Please make sure to add a comment for attribution, and if there is a license to verify it is compatible with BSD-3.

microsetta_interface/templates/sample_results.jinja2
```diff
@@ -0,0 +1,104 @@
+{% extends "sitebase.jinja2" %}
+{% set page_title = "Sample Results" %}
+{% set show_breadcrumbs = True %}
+{% set show_logout = True %}
+
+{% block head %}
+  <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.10.21/css/jquery.dataTables.css" />
```
wasade, line 7: should all these sources come from `static/`?

microsetta_interface/templates/sample_results.jinja2
```diff
@@ -0,0 +1,104 @@
...
+    let taxonomy = "{{taxonomy}}";
+    let metric = "{{alpha_metric}}";
+    let sampleId = "{{sample.sample_id}}";
+    var opt = {"renderer": "canvas", "actions": true};  /* Options for the Vega embedding */
+
+    function createDataTable() {
+        return function() {
+            $('#taxonomyTable').DataTable({
+                ajax: {
+                    url: "{{public_endpoint}}/taxonomy/present/single/" + taxonomy + "/" + sampleId
```
wasade, line 26: Barcodes in Microsetta do not contain the Qiita study prefix. I'm not seeing anywhere that the `10317.` is being prepended?

microsetta_interface/templates/sample_results.jinja2
```diff
@@ -0,0 +1,104 @@
...
+</div>
+<div class="row">
+  <div class="column" style="text-align: center">
+    <h3> What is in your sample? </h3>
+    <div style="padding:10px">
+      <table id="taxonomyTable" class="display" style="width:100%">
+        <thead>
+        <tr>
+          <th>Relative Abundance</th>
+          <th>Family</th>
```
wasade, line 91: No domain, phylum, class, order?

microsetta_interface/templates/source.jinja2
```diff
@@ -133,6 +136,9 @@
                 <div class="col-sm">
                     {{sample.sample_site |e}}
                 </div>
+                <div class="col-sm">
+                    <a class="btn btn-outline-info" href="/accounts/{{account_id}}/sources/{{source_id}}/samples/{{sample.sample_id}}/results">View</a>
```
wasade, line 140: Is there a check to make sure there are results for the sample?

microsetta_interface/static/css/4_column_flex.css
gwarmstrong, line 1:
```suggestion
/* https://www.w3schools.com/css/tryit.asp?filename=trycss_mediaqueries_flex */
  * {
```

microsetta_interface/templates/sample_results.jinja2
gwarmstrong: That info comes in from the API but left it out for brevity. Can include if you like.

wasade: Yes, thanks. Participants will be asking about the other taxonomic levels so best to include domain -> genus

microsetta_interface/templates/source.jinja2
gwarmstrong: Seemed to have overlooked exposing and endpoint for checking if sample results exist in `microsetta_public_api`. Can wrap it in with a new feature. Something like `/results-api/diversity/alpha/exists?sample_id=some_id`

gwarmstrong: And likely a similar feature for taxonomy

microsetta_interface/templates/sample_results.jinja2
gwarmstrong: Yeah... Also we did not pass the correct variable to `sampleId`. Rought draft 😅

microsetta_interface/templates/sample_results.jinja2
gwarmstrong: What are the pros/cons of that vs. using a CDN?

microsetta_interface/templates/source.jinja2
wasade: okay. What happens if a result has an alpha diversity value but not a taxonomy value? (shouldn't happen, but technically a possible outcome)

microsetta_interface/templates/sample_results.jinja2
wasade: Note sure TBH. @dhakim87?

microsetta_interface/server_config.json
gwarmstrong: Sure.
If we wanted to use that the options I see would be something like:
1) Choose one of the available taxonomy resources. E.g., just grab the first one in the list, or lexicographical minimum of them. Caveat: Does not guarantee we will use the specific resource we would like, if there is more than one. Seems a little weird.
2) Give the user a select picker to choose from. But then this has the caveats:
    a) they either see whatever variable names you gave the resources in the microsetta-public-api config
    b) OR we have to manually remap the variable names you gave them to things that appear to make more sense to the user
    c) This option doesn't really make sense if there is only one taxonomy resource corresponding to the given sample
3) You manually configure the taxonomy resource that the users see on their 16S samples, as above. Caveat: is not as flexible in the long run.

wasade: Okay thanks. Let's continue fwd with (3) and revise as the needs of the page grow

microsetta_interface/templates/sample_results.jinja2
gwarmstrong: Updated in https://github.com/biocore/microsetta-interface/pull/11/commits/fd7d375a6ac31e7cf08bace8f770b65c53089b37

microsetta_interface/templates/sample_results.jinja2
gwarmstrong: Updated in https://github.com/biocore/microsetta-interface/pull/11/commits/fd7d375a6ac31e7cf08bace8f770b65c53089b37

microsetta_interface/templates/source.jinja2
gwarmstrong: https://github.com/biocore/microsetta-interface/pull/11/commits/f00f8c492500290f2a0afdc5b6734cc39bf37b63
Only populates the button if the data have existing alpha diversity and taxonomy
Screen Shot 2020-07-27 at 9 52 32 PM

microsetta_interface/templates/sample_results.jinja2
gwarmstrong: Updated in https://github.com/biocore/microsetta-interface/pull/11/commits/07ad4628e717c96ae820900a0f2783a6ad09f844

PR #12, opened 2020-08-07
Emperor Plots

dhakim87:
This is a prototype that displays emperor plots in a new page accessible through the admin toolbar.  You can point it towards test data hosted on staging by modifying your server_config.json.  

Some of the emperor functionality doesn't work with the currently hosted test datasets, this is mostly due to nulls in some of the metadata fields, once this is resolved we can reevaluate to ensure that all the functionality we care about is implemented.  

Several minor modifications to Emperor were required to split out a .js file containing the Emperor initialization from the index.html file that we tend to think of as the root of an emperor installation.  These changes consist of splitting out the emperor_loader.js file, adding a callback for when the emperor control is ready, and specifying the root directory of the emperor installation (as the .html files can no longer reference these files with a fixed relative path).  Several other potential modifications could make embedding simpler: 

- Specifying a particular sample ID as the active/highlighted sample
- Restricting all changes to the html to fall within the #emperor-notebook-0x559558fb div
- Exposing the rest of the Emperor configuration support at the emperor_loader layer

Because these modifications are necessary to embed emperor within our own pages, we need to think very carefully about how we can stay up to date with new Emperor releases.  Ideally some/all of these changes are shifted into the main Emperor repository, simplifying our lives when new versions are available.


wasade:
Thanks, @dhakim87! Re-deploying to address the null item in just a second

dhakim87:
Screen Shot 2020-08-07 at 2 49 36 PM

wasade:
Would it make sense to have the emperor repo as a submodule?

dhakim87:
I'd prefer that to hosting a static version of it the way we do for bootstrap or jquery.  I think if we still intend to update Emperor and add new features, hosting it as a submodule lets us get those updates automatically.  Major requirement is that it must always be pointing to a stable version - any non backwards compatible changes to the Emperor api would result in a broken widget on the live site otherwise

microsetta_interface/static/vendor/emperor/templates/logic-template.html
```diff
@@ -0,0 +1,155 @@
+<div id='{{ plot_id }}' style="position: relative; width:{{ width }}; height:{{ height }};">
```
ElDeveloper, line 1: You don't need this file.

microsetta_interface/static/vendor/emperor/templates/jupyter-template.html
```diff
@@ -0,0 +1,3 @@
+{# Load templates dynamically, so we can figure out the path on the fly #}
```
ElDeveloper, line 1: You don't need this file.

microsetta_interface/static/vendor/emperor/templates/style-template.html
```diff
@@ -0,0 +1,15 @@
+<script type="text/javascript">
```
ElDeveloper, line 1: You don't need this file.

microsetta_interface/static/vendor/emperor/templates/logic-template.html
dhakim87: I think the messiest part about maintaining this integration is figuring out exactly what our api to emperor should be.  I'm worried that if updating to the newest version of emperor is any more complex than unzipping a file and drag dropping it into /static/vendor/emperor, then we're either going to run into constant problems on updates or we're going to be stuck using old versions for long periods of time.  

It seems we have a script that is capable of generating such a zip file (save for the changes I made to split index.html, which I'd also like to offload somehow), if we're worried about keeping this file and the other templates you noted then we should remove them as part of the configuration of the script that builds our installs?

wasade:
Thanks, @dhakim87 and @ElDeveloper. Is there an ETA for when these items can be sorted out so we can advance this visual into the participant result page?

microsetta_interface/static/vendor/emperor/templates/logic-template.html
ElDeveloper: Having a small script that can run the import would probably not be a bad idea. This would likely be not too bad to handle using the Emperor Python API. This way dependencies are handled via pip/conda.

wasade:
@ElDeveloper and @dhakim87, just a bump on this. It would be _awesome_ to get this integrated into the participant results pages!!

PR #31, opened 2020-09-11
Sample Notes Don't Say None

### LINKED ISSUE #8: Auto-delete "None" Upon Clicking
In the sample information page, "None" in the "Notes" section does not auto-delete when a participant clicks into it

13 - Edit Sample Information

dhakim87:
Fixes sample_notes to not display None to the user when no notes are set.  Instead we show a placeholder asking for any additional information they wish to provide

Fix #8

### DIFF
--- a/microsetta_interface/templates/sample.jinja2
+++ b/microsetta_interface/templates/sample.jinja2
@@ -134,7 +134,7 @@
           <div class="form-group">
             <label for="sample_notes" id="sample_notes_label">Notes</label>
             <div class="col-sm-10">
-                <textarea class="form-control" id="sample_notes" name="sample_notes" rows=3>{{ sample.sample_notes |e}}</textarea>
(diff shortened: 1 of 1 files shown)


microsetta_interface/templates/sample.jinja2
```diff
@@ -134,7 +134,7 @@
           <div class="form-group">
             <label for="sample_notes" id="sample_notes_label">Notes</label>
             <div class="col-sm-10">
-                <textarea class="form-control" id="sample_notes" name="sample_notes" rows=3>{{ sample.sample_notes |e}}</textarea>
+                <textarea class="form-control" id="sample_notes" name="sample_notes" placeholder="Is there any additional information you think we should know about this sample?  Time of day, special diet, recent antibiotics, collection issues, etc." rows=3>{{ sample.sample_notes if sample.sample_notes is not none else '' |e}}</textarea>
```
wasade, line 137: Time of day would be redundant with the collection time stamp. Recent abx is part of the survey. What about just "(optional) Is there else about this sample that you would like to add?"

PR #40, opened 2020-09-23
TEENY change to make login layout responsive

AmandaBirmingham:
This change ensures that the login page elements stack on a small or medium sized screen (iphone, ipad)
Screen Shot 2020-09-18 at 10 01 52 AM


dhakim87 (approved):
👍 Ahh, so that's how you're supposed to do it!  Now my source card changes just look stupid in comparison

PR #44, opened 2020-09-24
fixes #39 (inconsistent fonts on page)

### LINKED ISSUE #39: Multiple fonts are used on the account overview page
It looks like a serif and non-serif font. The inconsistency is a little unexpected and I think unnecessary
Screen Shot 2020-09-22 at 3 57 41 PM

AmandaBirmingham:
Turns out the issue is a result of taking zeplin at its word, css-wise.  Unfortunately, its auto-generated, overly-specified boilerplate css doesn't include backup fonts, thus defaulting to Times New Roman if one doesn't have Muli ... it also references css variables that we don't have defined :(. I've tried to excise the bits that are either breaking things or unused in the page.  Now looks like this:
Screen Shot 2020-09-24 at 3 56 21 PM

### DIFF
--- a/microsetta_interface/templates/account_overview.jinja2
+++ b/microsetta_interface/templates/account_overview.jinja2
@@ -25,8 +25,8 @@
   min-width: 277.5px;
 }
 .info-card {
-  border: solid 1px var(--forms_cards_outsideborder);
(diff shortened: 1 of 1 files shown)


wasade (approved):

wasade:
Thanks!!

PR #49, opened 2020-09-26
WIP Show vioscreen pdf to user

dhakim87:
Notes:

We worry about pushing retrieval from vioscreen out to celery, but then when talking from interface to private api, we do everything synchronously.  (But then, our async requests are waited on, and thus become synchronous... ugh)

Unclear if we want embedding or just pdf downloads

Should be worked into the site map

Possible issue with survey identifiers referring to both primary surveys and vioscreen data.


microsetta_interface/routes.yaml
```diff
@@ -494,6 +494,45 @@ paths:
...
+      responses:
+        '200':
+          description: Top Food Report pdf
+          content:
+            text/html:
+              schema:
+                type: string
+        '302':
+          description: Redirecting to necessary action
+
```
dhakim87, line 519: Add a 404 option here

microsetta_interface/routes.yaml
```diff
@@ -494,6 +494,45 @@ paths:
...
+      summary: Generate vioscreen top food report
+      description:  Generate vioscreen top food report
+      parameters:
+        - $ref: '#/components/parameters/account_id'
+        - $ref: '#/components/parameters/source_id'
+        - $ref: '#/components/parameters/survey_id'
+      responses:
+        '200':
+          description: Top Food Report pdf
+
```
dhakim87, line 535: Add a 404 option here?

microsetta_interface/templates/embedded_pdf.jinja2
```diff
@@ -0,0 +1,27 @@
...
+#app {
+    margin: 0;
+    height: 100%;
+}
+</style>
+{% endblock %}
+{% block content %}
+<!-- TODO: Needs an animated loading bar -->
+<div class="alert alert-info" role="alert">
+This report is generated by a third party.  Any identifying personal information: age, height, weight, etc was not shared in the generation of this report.
```
wasade, line 24: Can this be: "This report is generated by a third party without information regarding age, height or weight."

wasade:
@dhakim87 just one comment, and it looks like there are two items on 404s?

PR #53, opened 2020-10-16
Account Details Basic UK Support

### LINKED ISSUE #52: Support for International Addresses
In anticipation of our U.K. campaign launching soon--

- Give option to users to choose "U.K." as a country--although I think the current input field for this is hidden?  This is line 152 in the account_details.jinja2 template:
`<input id="country_code" name="country_code" type="hidden" value="{{account.address.country_code |e}}"/>`

If there's a way to make the page dynamic and change based on the country selection, the following should be updated:
- City changes to "Town"
- State updates to "County", different options repopulate the dropdown/changes to a text input field/if not essential, it can be hidden
- Zip Code changes "Postcode"

At minimum, we can also simply change some of the labels to be more inclusive of different address terminology:
- "City/Ciudad" is changed to "City/Town/Ciudad"
- "State/Estado" would still need the dropdown to repopulate/change to a text input field/be hidden if not essential
- "Zip Code/Codigo Postal" is changed to "Zip Code/Postcode/Codigo Postal"

dhakim87:
Fix #52  

Copies over ruleset.js from microsetta-admin (We need a common library!) Splits out country/state information from account_details.jinja2 into two new files, us_states.js copies over the state information we had, country_state_list is my fork of a github gist linking countries to states.  No state codes unfortunately, but it should be extensible to most anywhere in the world.  Uses ruleset to update text and options on the account_details page when country code is chosen


wasade (approved):
Thanks!! I agree with the centralization piece. Definitely important. Discuss during our meeting tomorrow?

PR #57, opened 2020-10-22
Missing container div

### LINKED ISSUE #56: Container class forcing elements to have different margins
What the interface looks like currently:
screencapture-microsetta-rest-ucsd-edu-accounts-a2b8a36c-7474-4c51-bd8c-72a83ec93485-2020-10-21-20_09_39

Some elements have a `<div class="container"></div>`, and the container element has a "margin-left: auto" style in the CSS, which is forcing those elements to be pushed over to the right.  Suggestion is to manually force the container class to have a margin-left of zero in the css file by adding in the following:
`
.container {
margin-left: 0;
}`

But that also leads to a mis-alignment of the sample table columns (see below), which I'm not too sure how to fix:
Screen Shot 2020-10-21 at 8 28 53 PM

dhakim87:
Fix #56 
Wrapped header columns in a container div so they'd be offset the same as the list group items

### DIFF
--- a/microsetta_interface/templates/source.jinja2
+++ b/microsetta_interface/templates/source.jinja2
@@ -140,25 +140,27 @@
     {% if samples|length > 0 %}
     Samples (click on a barcode to provide collection information):
     <div class="list-group">
-        <div class="row">
-            <div class="col-sm">
-                <i>Barcode</i>
+        <div class="container">
+            <div class="row">
+                <div class="col-sm">
+                    <i>Barcode</i>
+                </div>
+                <div class="col-sm">
+                    <i>Collection date</i>
+                </div>
+                {% if is_human %}
+                <div class="col-sm">
+                    <i>Sample type</i>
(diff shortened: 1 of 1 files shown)


microsetta_interface/templates/source.jinja2
```diff
@@ -140,25 +140,27 @@
...
-            {% if is_human %}
-            <div class="col-sm">
-                <i>Sample type</i>
-            </div>
-            <div class="col-sm">
-                <i>Results</i>
-            </div>
-            <div class="col-sm">
-                <i>Sample-specific survey</i>
+        <div class="container">
```
wasade, line 143: Is this paired with a closing `</div>`?

dhakim87: Yes, line 143 maps to line 164.  Github's diff isn't super clean for showing tabbing in the inner section.

PR #63, opened 2020-12-09
WIP: Cool Results Page

dhakim87:
***Work In Progress***, don't merge this yet!  

Skeleton for new results page with events linked, showing where to fire off ajax requests and what to call to update controls.  I need to know what words to put on each of the four tabs, what kind of plot(s) to show in that tab, and what urls to hit on public api to grab the data for that tab.


dhakim87:
Screen Shot 2020-12-08 at 7 31 26 PM

wasade:
Thanks, @dhakim87! This is looking pretty good. I had some follow up with our comms team, and they were suggesting opening with some fairly high level text like the stuff below. Some of which may require some additional bits on the public API (cc @gwarmstrong), like the rare/unique bit, which is part of what I'm trying to knock out this week. 

I'll email the mockups of the other pages, which are at the moment mostly just the content from the current results page with the exception of taxa ranks and some more user-guided bits on the existing alpha diversity plot. 

@gwarmstrong, with the nearest neighbors piece, I think right now we could do some basic text stuff like this (i.e., the person you're most similar to is XXX years old, and eats sugary sweets with YYY frequency) but I think under the diversity similarities section. I'll summarize a bit more on slack today. It would be advantageous for the nearest neighbors bits to be dataset aware, so that we could (for example) depict what mammal someone is most similar too.

```text
{% if has_unique_genera %}
Your {{ SAMPLE_TYPE }} sample was _super_ interesting! Out of {{ LARGE_NUMBER_OF_PEOPLE }} people we’ve sequenced in the {{ COUNTRY_OF_RESIDENCE }}, you’re the only one with {{ UNIQUE_GENERA }} microbes!” 
{% else %}
Your {{ SAMPLE_TYPE }} sample was _super_ interesting! Out of {{ LARGE_NUMBER_OF_PEOPLE }} people we’ve sequenced in the {COUNTRY_OF_RESIDENCE}, you’re one of only {{ PEOPLE_WITH_GENERA }} people with {{ RARE_GENERA}} microbes!” 
{% endif %}

Research has shown microbiomes are more unique than our genomes — if you an identical twin, you do not have an identical microbiome [HREF -> LEY TWINS WORK]! Your sample is helping us to unravel this enormous diversity, and to better understand what it means in relation to health and lifestyle factors. 

For your results, we’ve compiled information about how diverse your sample is [LINK TO YOUR SAMPLE DIVERSITY], and ways you can compare yourself to others. We’ve also assessed how similar your microbiome is to {{ TOTAL_NUMBER_OF_SAMPLES }} other samples, which you can explore interactively [LINK Diversity similarities]. And last, but not least, we’ve applied cutting edge computational algorithms to assess what types of microbes are on your sample [LINK TO TAXONOMY]. 

Have fun exploring, and please make sure to check back as we’re continuing to expand on what we can say about your sample!
```

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -0,0 +1,145 @@
...
+        $("#similarity").text(similarityData["result"])
+    }
+    updateTaxonomy = function(taxonomyData){
+        //TODO
+        $("#taxonomy").text(taxonomyData["result"])
+    }
+
+    init = function(){
+        // TODO: Should data type options be determined by ajax call or
+        // passed into jinja2 template?
```
wasade, line 53: Probably ajax as this is coming from the public api. If a sample was only subject to one data type, then the pulldown shouldn't be selected

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -0,0 +1,145 @@
...
+        populate_select(
+            "#data_type",
+            {
+                "16s": "16S",
+                "shotgun": "Shotgun"
+            },
+            "16s"
+        );
+
+        // TODO: Should dataset options be determined by ajax call or
```
wasade, line 63: Probably ajax as this information is coming from the public api

gwarmstrong:
> nearest neighbors bits to be dataset aware

AFAIK the `neighbors` endpoint is dataset aware `GET
​/dataset​/{dataset}​/diversity​/beta​/{beta_metric}​/nearest`

> so that we could (for example) depict what mammal someone is most similar to

This seems like a slightly different problem than the neighbors endpoint supports. The `neighbors` endpoint currently supports "get the k most similar samples to a given sample in a given dataset", which is just a sort on the distance matrix column/row.

For the mammals bit, it seems we may want an additional feature that supports "get the k most similar samples in this dataset that meet a certain criteria", which is different because it would also involve searching for samples that meet the criteria.

wasade:
Correct, although it doesn't need to be expressed as a search but instead a filter prior to sorting. We need to expand slightly to support overall nearest neighbors, and a restricted nearest neighbors

gwarmstrong:
> Correct, although it doesn't need to be expressed as a search but instead a filter prior to sorting. We need to expand slightly to support overall nearest neighbors, and a restricted nearest neighbors

Moved to new issue: https://github.com/biocore/microsetta-public-api/issues/74

wasade:
Thanks!

PR #68, opened 2020-12-16
Merge dataset and data type select boxes, use ajax

dhakim87:
* Can treat both dataset and data type as a single variable in the UI.  
	* All values depend on Dataset (Data Type is merging with dataset)

Collapsed Dataset and Data Type into just Dataset.  Some hacky workarounds for jquery event handling while Emperor is loading in.  Ajax is used to fill dataset.


wasade (changes requested):
Seems fine, but same comment regarding defensive JS

PR #71, opened 2021-01-23
Delete Source - Interface Side

dhakim87:
Adds new accounts/acct_id/sources/src_id/remove route to remove a source (that has no samples associated with it)
Adds a button to bottom of source.jinja2 to remove.
Button is disabled and shows tooltip if source cannot be deleted due to existence of samples.  
Tooltip implemented with popper 1.11.0 (Need to upgrade to newest bootstrap to upgrade to popper v2)

See private api issue at https://github.com/biocore/microsetta-private-api/issues/288


microsetta_interface/templates/source.jinja2
```diff
@@ -256,5 +274,16 @@
         {% endfor %}
     </div>
     {% endif %}
+    <hr style="border-top: dotted 1px"/>
+
+    <div class="container">
+        <div class="row">
+            <div class="col-12 {% if samples|length > 0 %}tooltipper{% endif %}" data-title="Sources with one or more samples assigned cannot be deleted.  All samples above must be removed prior to deleting this source.">
+                <form name="delete_{{source_id}}_form" method="post" action="/accounts/{{ account_id }}/sources/{{ source_id }}/remove" onsubmit="return verifyDeleteSource('{{source_id}}');">
```
AmandaBirmingham, line 282: I suggest passing `source_name` rather than `source_id` to the javascript confirm function; see comment above.

microsetta_interface/templates/source.jinja2
```diff
@@ -81,6 +87,12 @@
             return window.confirm(confirmMsg);
         }
 
+        function verifyDeleteSource(sourceId){
+            let confirmMsg = "Deleting this source will discard all surveys associated with it. " +
```
AmandaBirmingham, line 91: We humans are just awful with deletion warnings--we tend to ignore them until after we hit confirm <smh>.  We might get slightly fewer mistaken deletions if we passed in the user-readable `source_name` and modified the first sentence of the message to something like 

> You are deleting source "`source_name`" and every survey associated with it!

AmandaBirmingham (commented):
Looks good and straightforward to me (well, except for the tooltip stuff which looks like it was painful to work out!) Small suggestion for delete message, not a must-have.

microsetta_interface/templates/source.jinja2
dhakim87: Done

microsetta_interface/templates/source.jinja2
dhakim87: Done

PR #75, opened 2021-02-09
Disables interface controls for disassociating samples and editing samples

dhakim87:
Interface didn't have support for using sample_locked flag in the past.  With split from sample_locked to sample_edit_locked and sample_remove_locked, this updates interface to disable editing controls or removal from sources as appropriate.  

May want to wait for popper.js to come in so we can add a tooltip to the remove button before merging.


wasade:
:+1: and agree on popper, thanks!

PR #79, opened 2021-02-24
Empress playground

gwarmstrong:
Depends on https://github.com/biocore/microsetta-public-api/pull/96

This PR adds a page to the admin site for viewing an empress visualization, similar to the emperor playground. Seen below:

Screen Shot 2021-02-24 at 10 09 55 AM

There are some withstanding CSS issues (that seemed relatively complex to solve, without changes to the CSS and sitebase), but otherwise, this should work nicely with the current development deployment of the public API.

The vendored empress files are taken directly from the empress 1.1.0 release.

`microsetta_interface/templates/empress-template.html` required some modifications to the `empress-template.html` from the release.


wasade:
Thanks, George! 

How will version updates to empress work? Should this be a submodule or install or some other mechanism to bring in the empress codebase?

Were issues opened up regarding modifying the state of empress at initialization, and do the mentioned CSS issues block usability for participants?

gwarmstrong:
> How will version updates to empress work? Should this be a submodule or install or some other mechanism to bring in the empress codebase?

As far as I can tell, a submodule could work for versioning, by checking out tags, e.g., [this SO post](https://stackoverflow.com/questions/38384670/github-branch-tag-how-to-get-a-specific-release-of-a-code). But that could also be a bit overkill because as far as I can tell, there is no way to just check out the web assets, so all of the python files etc, would be included too (maybe or maybe not desirable).  I don't think it would be terrible to write a script that downloads a specific version of empress and replaces the existing support files.


> Were issues opened up regarding modifying the state of empress at initialization

I actually think we can solve the issue somewhat succinctly without any modifications to empress, by invoking `click` methods on the side panel. See this line https://github.com/biocore/microsetta-interface/blob/5c0bd1f15c2707a8e168b6a6cd75f704c2f0a5a9/microsetta_interface/templates/empress-template.html#L211

> and do the mentioned CSS issues block usability for participants?

maybe not usability, but its certainly ugly. Some of the empress CSS modifies microsetta defaults. Notice how the "Administrator toolbar" text is smaller on the empress page. And microsetta CSS modifies default values depended on by empress, e.g., (not shown) the check marks in radio buttons are mis-aligned on microsetta empress, presumably due to some padding/margin value somewhere. Changing the order of the CSS imports only changes which things go wrong.

Screen Shot 2021-02-23 at 4 47 31 PM
Screen Shot 2021-02-23 at 4 47 36 PM

wasade:
Thanks, George. Having users click on something in an unknown interface will create a lot of confusion, and I worry about the CSS fighting as that can realistically play hell in unexpected ways (like responsiveness). I think these are important items to understand, however if they are likely to take months to fix, then we should consider placing this under an "Experimental results" tab for participants and provide details and set expectations 

cc @ElDeveloper @dhakim87

gwarmstrong:
> Thanks, George. Having users click on something in an unknown interface will create a lot of confusion

Right, so if we call the `click` method in javascript, or whatever else we need, the user does not have to configure anything to see our default view.

wasade:
Ah okay. Does this also highlight the individuals sample against the background dataset?

gwarmstrong:
Same idea applies, I believe. If I have information about a sample's features, I can add it as feature metadata to the empress object, and I believe I should I should be able to use javascript methods to set the feature metadata coloring to this category.

However, this is still pending an endpoint on public api. I could do a proof of concept with mock data at some point in the next few weekdays.

wasade:
Okay thanks! I think that would be super helpful to convey what users may be able to see with this

gwarmstrong:
> I worry about the CSS fighting as that can realistically play hell in unexpected ways (like responsiveness). I think these are important items to understand, however if they are likely to take months to fix, then we should consider placing this under an "Experimental results" tab for participants and provide details and set expectations

@ElDeveloper and I chatted, and it seems like having a standalone empress page that we embed as an iframe on the results page seems like potential solution for the CSS issues.

wasade:
Excellent!!

gwarmstrong:
@wasade The iframe solution worked! Just pushed some commits for it.

gwarmstrong:
@wasade 
Note that this now depends on https://github.com/biocore/microsetta-public-api/pull/96

Here is an example with feature metadata from results API showing features that are in a sample:
Screen Shot 2021-06-04 at 1 28 34 PM

wasade:
The code here is relevant for future efforts but I dont right now see a path to making this available reasonably to participants -- closing for now, please reopen if that assessment is incorrect

PR #83, opened 2021-03-02
Open results experimental

wasade:


dhakim87:
Yeah, this'll work.  I had thought you'd copy paste it and add a route, but I guess we can just undo this change when we're ready to go live anyway.

wasade:
oh i see so that we can also test on arbitrary samples if authenticated? Good call, i'll readd that

PR #100, opened 2021-03-09
Bolded and localized bacteria and archaea counts

### LINKED ISSUE #85: Numbers like the number of unique bacteria should stand out
...should be bold, and differentiate from what hyperlinks look like. Numbers should be localized with the correct separator (eg 1000 -> 1,000)

dhakim87:
Fixes #85


dhakim87:
Fix #85

PR #101, opened 2021-03-10
Tool tips, and text revisions suggested by @justinshaffer

wasade:
Updated text based on @justinshaffer's comments, added in two tool tips, one for "taxa" and one for "dataset", and slightly reorganized the text for the how you compare section

The coloring of the tooltip is not ideal. @b4collinsworth and I tried to use UCSD Grey but it was forcing a transparent background, and popper did not seem to want to honor `opacity: 1`. I screwed with it for about 30 minutes, then gave up, it's possibly related too [this](https://github.com/popperjs/popper-core/issues/625). For the time being, getting something in with content and we can revisit the exact feel later. 

An example screenshot is attached (note: my env was not sourcing from the results-api so the ad lib pieces are unfilled) 

Screen Shot 2021-03-10 at 11 49 22 AM


dhakim87 (approved):
👍

PR #102, opened 2021-03-11
What is in your sample revisions

dhakim87:
- [ ] Should provide a helpful link out to how to understand taxonomy

- [ ] Feature classification link should be to the latest QIIME 2 link, not old versions

- [ ] Needs guidance in reading the table

- [ ] We should provide a summary text of some commonly observed bacteria, and what can be safely said about them

- [X] Rank plot is limitedly described, unsure how to interpret. Needs axis labels

- [X] i suggest to replace 'In some cases, a sequence may uniquely match a database record, in other cases a sequence may match different records.' with 'In some cases, one of the DNA sequences we recover from your sample may uniquely match a database record, in other cases a sequence may match two or more different records.'
- [X] - i suggest to replace '...and distribution of ranks of those genera' with '...and relative abundance of those taxa in the entire Microsetta dataset'
- [X] - i suggest to add axes labels to the figure
- [X] - i suggest to explain what it means when a taxon has brackets around its name
- [X] - i suggest to add units to the relative abundance column (e.g., 'proportion of sequences')

### DIFF
--- a/microsetta_interface/templates/new_results_page.jinja2
+++ b/microsetta_interface/templates/new_results_page.jinja2
@@ -33,7 +33,8 @@
 <script type="text/javascript" charset="utf8" src="/static/vendor/js/jquery.dataTables.js"></script>
 <script src="/static/vendor/DataTables/Buttons-1.6.2/js/dataTables.buttons.min.js"></script>
 <script src="/static/vendor/DataTables/Buttons-1.6.2/js/buttons.html5.min.js"></script>
-<!-- plotly must precede emperor imports as well.   I'm getting the feeling
+<script src="/static/vendor/DataTables/PercentageBars-1.10.21/js/percentageBars.js"></script>
+    <!-- plotly must precede emperor imports as well.   I'm getting the feeling
      that emperor doesn't know how to play nice. -->
 <script src='https://cdn.plot.ly/plotly-latest.min.js'></script>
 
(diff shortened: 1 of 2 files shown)


microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -699,7 +708,7 @@
                       <table id="taxonomyTable" class="display" style="width:100%">
                         <thead>
                         <tr>
-                          <th>Relative Abundance</th>
+                          <th>Relative Abundance (Fraction of Sequences)</th>
```
wasade, line 711: Would it be easy to transform this to "% of your sample", and round to 2 sig digits?

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -494,9 +498,13 @@
             let layout = {
                 "title": "Ranked Relative Abundance",
                 "yaxis": {
+                    "title": "Common Genera",
```
wasade, line 501: "Common taxa"

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -494,9 +498,13 @@
...
                     "automargin": true,
                     "categoryorder": "array",
                     "categoryarray": topXOrdered
+                },
+                xaxis: {
+                    title: 'Rank'
```
wasade, line 507: "<-- less abundant    |     more abundant -->"

(unless that looks totally ugly)

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -494,9 +498,13 @@
             let layout = {
                 "title": "Ranked Relative Abundance",
```
wasade, line 499: "How abundant are our taxa compared to everyone else?"

wasade (changes requested):
Some minor text edit suggestions

PR #106, opened 2021-03-12
Tool tip on remove sample

wasade:
Adds a tool tip for removal of a sample when the removal is not possible. I cannot get the positioning perfect, it looks like it is aligning to the middle of the div rather than the button. Attempting to place it directly on the button did not immediately work. I think we should defer this for a future review of the tooltip aesthetics, and focus right now on tool tip content. 

Screen Shot 2021-03-11 at 4 55 03 PM


AmandaBirmingham (approved):
Seems straightforward :)

wasade:
Thanks!!

PR #120, opened 2021-04-09
Account Creation by Activation Code

dhakim87:
Lets user enter an Activation Code instead of a Kit ID to login.  

Probably shouldn't merge this until there is actually a way to distribute activation codes to users??

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -1314,14 +1314,37 @@ def post_generate_activation(body):
         raise Unauthorized()
 
     email = body["email"]
-    do_return, diagnostics, _ = ApiRequest.post(
-        "/admin/activation",
-        json={"emails": [email]}
-    )
 
-    if do_return:
-        return diagnostics
+    # Generate the activation code and update the list
+    if 'generate' in body or 'generate_send' in body:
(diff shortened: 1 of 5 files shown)


microsetta_interface/implementation.py
```diff
@@ -1203,6 +1205,20 @@ def get_ajax_list_kit_samples(kit_name):
     return flask.jsonify(result), code
 
 
+def get_ajax_check_activation_code(code, email):
+    response = requests.get(
+        ApiRequest.API_URL + '/can_activate',
+        auth=BearerAuth(session[TOKEN_KEY_NAME]),
+        verify=ApiRequest.CAfile,
+        params=ApiRequest.build_params({"email": email, "code": code}))
```
wasade, line 1213: I don't think it matters if the email address is sent, just that it's a valid code?

wasade: ...if the email is sent, then we would need to prompt the user to make sure they use the same email for which they received an activation code for

wasade (commented):
Question on emails, otherwise this looks good!

dhakim87:
This PR is now updated to allow sending activation emails one at a time through the admin toolbar as activation codes are generated.

PR #124, opened 2021-04-27
Survey Aesthetics

cassidysymons:
1) On new_participant.jinja2, there are two groups of changes:
1a) Not all fields that are required were actually shown as required to the user - I standardized that
1b) The asterisk for required fields was breaking onto a new line - I formatted it to be red and show up on the same line with cleaned-up visuals

2) On survey.jinja2:
2a) I paginated the form by field group
2b) Added buttons for Previous & Next Section
2c) Added a progress bar to the top of the screen
2d) Add a one-time confirm on submit since responses can't be edited
2e) Made some font weight and size changes to decrease the wall-of-text effect

(note, depends on biocore/microsetta-private-api#315)

### DIFF
--- a/microsetta_interface/templates/new_participant.jinja2
+++ b/microsetta_interface/templates/new_participant.jinja2
@@ -156,10 +156,10 @@
            <p><input value="Yes" type="checkbox" name="consent_parent" id="consent_parent" required data-rule-required="true"> {{ tl['TEXT_I_HAVE_READ_PARENT'] |e}} <span class="required-field">*</span></p>
     </div>
     <table>
-        <tr><td><label for='participant_name'>{{ tl['PARTICIPANT_NAME'] |e}}</label></td><td><input tabindex="2" type="text" name="participant_name" id="participant_name" required data-rule-required="true"> <div class="red right required-field">*</div></td></tr>
-        <tr><td>{{ tl['PARTICIPANT_EMAIL'] |e}}</td><td><input tabindex="3" type="email"  name="participant_email" id="participant_email" required data-rule-required="true"> <div class="red right required-field" required data-rule-required="true">*</div></td></tr>
-        <tr><td>{{ tl['PARTICIPANT_PARENT_1'] |e}}</td><td><input tabindex="7" type="text" name="parent_1_name" id="0-6-form_parent_1_name" required data-rule-required="true"> <div class="red right required-field">*</div></td></tr>
-        <tr><td>{{ tl['PARTICIPANT_PARENT_2'] |e}}</td><td><input tabindex="8" type="text" name="parent_2_name" id="0-6-form_parent_2_name" required data-rule-required="true"> <div class="red right required-field">*</div></td></tr>
+        <tr><td><label for='participant_name'>{{ tl['PARTICIPANT_NAME'] |e}}</label></td><td><input tabindex="2" type="text" name="participant_name" id="participant_name" required data-rule-required="true"> <span class="required-field">*</span></td></tr>
+        <tr><td>{{ tl['PARTICIPANT_EMAIL'] |e}}</td><td><input tabindex="3" type="email"  name="participant_email" id="participant_email" required data-rule-required="true"> <span class="required-field">*</span></td></tr>
+        <tr><td>{{ tl['PARTICIPANT_PARENT_1'] |e}}</td><td><input tabindex="7" type="text" name="parent_1_name" id="0-6-form_parent_1_name" required data-rule-required="true"> <span class="required-field">*</span></td></tr>
+        <tr><td>{{ tl['PARTICIPANT_PARENT_2'] |e}}</td><td><input tabindex="8" type="text" name="parent_2_name" id="0-6-form_parent_2_name" required data-rule-required="true"> <span class="required-field">*</span></td></tr>
         <tr><td>{{ tl['PARTICIPANT_DECEASED_PARENTS'] |e}}</td><td><input  tabindex="9" type="checkbox" name="deceased_parent" id="deceased_parent" value="true" onclick="relax_parents(this)"></td></tr>
         <tr><td><input type="submit" id="0-6-submit_button" value="I Accept"></td><td></td></tr>
     </table>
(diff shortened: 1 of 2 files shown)


microsetta_interface/templates/survey.jinja2
```diff
@@ -5,11 +5,60 @@
...
+            font-weight: bold;
+        }
+    </style>
     <script>
         var result_txt = "";
         var error_txt = "";
+        var form_submitted = 0;
 
         function postSurvey() {
+            fieldsetObject = document.getElementsByTagName("fieldset");
```
wasade, line 52: Could this be declared with either `let` or `var`? Same comment for the declarations in the subsequent functions

microsetta_interface/templates/survey.jinja2
```diff
@@ -94,10 +145,68 @@
             type: "submit",
             validateBeforeSubmit: true,
             onSubmit: function(){
-                $("input").prop('disabled', true);
-                return postSurvey();
+                if(form_submitted == 0) {
+                    confirm_submit = confirm("Please note: Once you've submitted this survey, you'll be unable to edit your answers. If you'd like to review or change any of your responses, please press the Cancel button. Otherwise, press OK to submit your responses.");
```
wasade, line 149: Could `confirm_submit` be declared within this function?

microsetta_interface/templates/survey.jinja2
```diff
@@ -94,10 +145,68 @@
...
+
+        function paginateForm() {
+            fieldsetObject = document.getElementsByTagName("fieldset");
+
+            if(fieldsetObject.length <= 1) {
+                document.getElementById("survey-nav-button-next").style.display = "none";
+            } else {
+                document.getElementById("form-progress").style.display = "block";
+                document.getElementById("form-progress").innerHTML = "Step 1 of " + fieldsetObject.length + " - 0% Complete";
+                for(i=1; i<fieldsetObject.length; i++) {
```
wasade, line 172:
```suggestion
                for(let i=1; i<fieldsetObject.length; i++) {
```

wasade (changes requested):
Thanks, @cassidysymons! Two general requests, can variables be declared for use relative to their scoping needs? And second, would it be possible to use strict equality (`===`) on conditionals?

wasade:
From interactive review:

The confirm window unexpectedly shows only once. I selected "Submit survey", then clicked "cancel", and then reclicked "submit survey". The second "submit survey" click proceeded without the confirmation window.

Would it be possible to default the selection as `Unspecified` on the radio buttons and if easy, to do so for the dropdowns? 

Any particular reason the alcohol levels have a pulldown but the other selections don't (see attached screenshot)?

Screen Shot 2021-04-27 at 9 01 09 AM

cassidysymons:
> The confirm window unexpectedly shows only once. I selected "Submit survey", then clicked "cancel", and then reclicked "submit survey". The second "submit survey" click proceeded without the confirmation window.

That was an intentional decision to make it fairly unobtrusive - essentially, just a one-time "Hey, you can't edit this, do you want to go back and check it?" If you think it will be confusing to users, I can change it to continue showing up until the users hits OK.

> Would it be possible to default the selection as Unspecified on the radio buttons and if easy, to do so for the dropdowns?

For some reason, the createDefaultObject function in VFG doesn't seem to actually populate default values. I see two options here:
1) Add a custom JavaScript function after the form is generated to walk through the form and choose the "Unspecified" option for any radio button and dropdown. My main concern is that if people see a question already answered, they might be less likely to change it, as there's a behavioral bias towards leaving default answers in place.
2) Eliminate the "Unspecified" option altogether since people have the implicit option to just not answer any given question. However, I'm unsure if this would be something that would require approval and/or introduce any sort of statistical issues in new surveys vs old surveys?

> Any particular reason the alcohol levels have a pulldown but the other selections don't (see attached screenshot)?

That question has seven possible answers, so it falls just outside of the criteria for changing to a radio button. We can move that limit around however we want, choosing six was fairly arbitrary in that it seemed like a manageable number of options.

wasade:
Thanks! For the confirmation window, that makes total sense, do you think it would be nice to note in the window this message will only appear once?

For Unspecified, I like option 2 except for the situation where a person selects a response and then decides they don't want to answer the question anymore. Is there a non terrible way to allow someone to unselect a radio button while also eliminating unspecified?

I agree with the behavior bias concern on 1. 

Ah I see where the limit occurs now, thanks. Would it be possible to bump it +1? I think that gets the alcohol question and one other. It does wrap around on small form factor devices, but that's true at six too

microsetta_interface/templates/new_participant.jinja2
```diff
@@ -118,6 +118,11 @@
   width: 80%;
   margin: 0 auto;
 }
+
+.required-field {
```
dhakim87, line 122: Looks like a bunch of the spans use class="red right required-field", I'm guessing class red just sets color: red.  Should we bundle the definition of right into here also, then we can just set class="required-field" everywhere and drop the red and right classes from all the spans?

cassidysymons: The "red" and "right" classes were already in the code but didn't appear to be doing anything. I added the required-field class to handle the aesthetic changes I was making, but left red and right in place in case there was some legacy reason for them that I was unaware of.

If there's no reason to keep them, I'd be happy to clean them all out and just keep required-field.

dhakim87: Ahh, so spans are required-field, divs are red right required-field?  I think my question is can we drop the red class from the divs (assuming the right class does something else we want that isn't covered)?

cassidysymons: As far as I can tell, both the red and right classes can be dropped from the divs. Neither class is defined anywhere in the CSS and I don't see any JavaScript that uses them. If neither you nor @wasade know of any reasons to keep them, I can drop both and just leave required-field in place.

wasade:
No preference on this

microsetta_interface/templates/new_participant.jinja2
dhakim87: I'd say drop them then, if nothing shifts around or changes color, we're in the clear

dhakim87 (approved):
This is a great change, the pagination code is very clear and the css changes on the vue elements look good as well.

microsetta_interface/templates/survey.jinja2
```diff
@@ -5,11 +5,60 @@
     <script type="text/javascript" src="/static/vendor/js/vue-2.5.17.min.js"></script>
     <script type="text/javascript" src="/static/vendor/vue-form-generator-2.3.4/vfg.js"></script>
     <link rel="stylesheet" type="text/css" href="/static/vendor/vue-form-generator-2.3.4/vfg.css">
+    <style type="text/css">
+        .vue-form-generator .field-radios .radio-list label {
```
dhakim87, line 9: My understanding is that this catches labels in groups of radio buttons, similar style be set for groups of checkboxes?  

(See Do you eat a specialized diet ? (select all that apply), Step 2 of 7 in the Microsetta Participant Survey)

Rough guess on the css selector is:
.vue-form-generator .field-checklist .dropList .list-row label

Could be a separate change entirely though.

microsetta_interface/templates/new_participant.jinja2
cassidysymons: I just dropped all of the references to the red and right classes and standardized all of the asterisks to <span class="required-field">. I double-checked after the change and nothing was altered visually, either in color or positioning.

microsetta_interface/templates/survey.jinja2
cassidysymons: Are you thinking you'd want to have the checkbox options span the entire width of the screen (similar to radio buttons), rather than in a one-per-line dropdown?

cassidysymons:
I updated the variable scoping and switched to strict equality conditionals - let me know if I missed anything on either of those fronts.

I also added a line to the confirmation box saying that it will only appear once.

As far as the Unspecified side of things, there's really no simple way to deselect a radio button once the user has selected it. It's possible to do via JavaScript, so we could theoretically add a button to each field to clear selections, but that's neither cleaner for the user, nor from the standpoint of code maintinability.

I can definitely see the issue in users not having a way to deselect a field after answering, whether due to erroneously clicking or deciding they prefer not to answer. Let me think about this topic a bit and circle back to it later, I feel like there's some sort of middle ground in making the "Unspecified" option both more useful and more graceful but I'm not seeing it at the moment.

wasade:
Thanks, @cassidysymons!! Sounds good on the unspecified item -- let's table it for now. I'll re-review this evening

microsetta_interface/templates/survey.jinja2
wasade: @dhakim87 is this a blocking item or can this be sorted out later?

wasade (approved):

wasade:
I'm +1, will merge after hearing on the open comment with @dhakim87

PR #131, opened 2021-05-05
Adds messages.pot and first usages of gettext

dhakim87:
First pass over implementation.py to add support for gettext and properly format dates.  

Added a super basic babel config, will need to add to it to support javascript, but should work for python and jinja2.  

To regenerate messages.pot (after you add gettext calls to python or jinja2 templates):
cd to root directory of checkout
pybabel extract -F babel.cfg -o messages.pot .

To check if your new strings made it past the extract call, 
cat messages.pot and look for your strings.

### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -1,0 +1,101 @@
+# Translations template for PROJECT.
+# Copyright (C) 2021 ORGANIZATION
+# This file is distributed under the same license as the PROJECT project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PROJECT VERSION\n"
+"Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
+"POT-Creation-Date: 2021-05-04 17:09-0700\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=utf-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Babel 2.9.0\n"
+
+#: microsetta_interface/implementation.py:379
+msgid "Unable to validate the kit name; please reload the page."
+msgstr ""
+
+#: microsetta_interface/implementation.py:398
+msgid ""
+"The provided kit id is not valid or has already been used; please re-"
+"check your entry."
+msgstr ""
+
+#: microsetta_interface/implementation.py:1012
+msgid "Blood (skin prick)"
+msgstr ""
+
+#: microsetta_interface/implementation.py:1013
+msgid "Saliva"
+msgstr ""
+
+#: microsetta_interface/implementation.py:1014
+msgid "Stool"
+msgstr ""
+
+#: microsetta_interface/implementation.py:1015
+msgid "Mouth"
+msgstr ""
+
+#: microsetta_interface/implementation.py:1016
+msgid "Nares"
+msgstr ""
+
+#: microsetta_interface/implementation.py:1017
+msgid "Nasal mucus"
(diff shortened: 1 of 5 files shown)


messages.pot
```diff
@@ -0,0 +1,101 @@
...
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Babel 2.9.0\n"
+
+#: microsetta_interface/implementation.py:379
+msgid "Unable to validate the kit name; please reload the page."
+msgstr ""
+
+#: microsetta_interface/implementation.py:398
+msgid ""
+"The provided kit id is not valid or has already been used; please re-"
```
wasade, line 26: is it weird this is on a newline?

microsetta_interface/implementation.py
```diff
@@ -929,7 +936,12 @@ def get_source(*, account_id=None, source_id=None):
             needs_assignment = True
         else:
             dt = datetime.fromisoformat(sample['sample_datetime'])
-            sample['sample_datetime'] = dt.strftime("%b-%d-%Y %-I:%M %p")
+            # sample['sample_datetime'] = dt.strftime("%b-%d-%Y %-I:%M %p")
+            # rebase=True - show in user's locale, rebase=False, UTC (I think?)
```
wasade, line 940: I believe this is `BABEL_DEFAULT_TIMEZONE` which is set to `PT` right now

microsetta_interface/implementation.py
```diff
@@ -995,15 +1007,37 @@ def get_update_sample(*, account_id=None, source_id=None, sample_id=None):
                         "Nares", "Nasal mucus", "Right hand", "Left hand",
                         "Forehead", "Torso", "Right leg", "Left leg",
                         "Vaginal mucus", "Tears", "Ear wax", "Hair", "Fur"]
+        # babel scraping doesn't understand anything but constant strings.
+        sample_site_translations = [
```
wasade, line 1011: Is that true though? Does the following not work?

```python
sample_site_translations = [gettext(s) for s in sample_sites]
```

...or is pybabel walking the AST to find `gettext` calls?

messages.pot
dhakim87: I think it likes breaking at 80 characters due to the defaults in babel.cfg, so it saw the - and went for it.

microsetta_interface/implementation.py
dhakim87: rebase=False will show in whatever the BABEL_DEFAULT_TIMEZONE is yeah.  If you set it to PT somewhere that's fine.

microsetta_interface/implementation.py
dhakim87: It does not!  At least not when I tried it.  That was my first try.

wasade: okay cool cool

messages.pot
wasade: cool cool

microsetta_interface/server.py
```diff
@@ -77,6 +79,42 @@ def run(app):
     )
 
 
+@babel.localeselector
+def get_locale():
```
wasade, line 83: This needs to either move below the definition of `babel` on line 119, or lines 118-119 need to be moved prior to the use of `babel.localeselector`

microsetta_interface/server.py
```diff
@@ -77,6 +79,42 @@ def run(app):
...
+    #  all the flask-babel examples seem to use this 'user' pattern, so
+    #  may be standard practice to leave it exactly as is in their example.
+    user = getattr(g, 'user', None)
+    if user is not None:
+        return user.locale
+
+    # TODO: We update this as we add support for new languages
+    return request.accept_languages.best_match(['en', 'es'])
+
+@babel.timezoneselector
```
wasade, line 112: Same here

microsetta_interface/implementation.py
wasade: kk

microsetta_interface/server.py
dhakim87: Oh wow, that syntax is super weird.  I thought that babel wanted the module, but I guess you're right

PR #135, opened 2021-05-10
Image localization

dhakim87:
Uses babel updates in jinja2 templates to point towards correctly localized image files.  
Assumes we will translate en_us to es_mx for our mexico deployment, and copy over translated images to that folder.  

Only the bristol stool chart seems to be used within surveys, which is the only one I couldn't do this way.  I added a separate PR to microsetta-private-api to update the sql on that side, but we may want to build more localization support on the private api side before merging that, or merge it to somewhere other than master.


microsetta_interface/templates/home.jinja2
```diff
@@ -19,16 +19,16 @@
 {% block content %}
     <div class="row">
       <div class="col-12 col-lg-8 text-center how_it_works">
-          <img src="/static/img/how-it-works.svg" class="img-fluid" />
+          <img src="/static/img/{{ _("en_us") }}/how-it-works.svg" class="img-fluid" />
```
wasade, line 22: That's pleasant

wasade (approved):

PR #143, opened 2021-05-12
Fix Dependabot Alert

dhakim87:
Upgraded emperor's dependency: underscore.js to umd version 1.13.1 in response to the dependabot alert on github


wasade:
cc @antgonza

wasade:
so i think this is fine, but @dhakim87 out of curiosity was any different type of behavior observed with emperor?

PR #144, opened 2021-05-18
Draft es_MX

wasade:
Draft es_MX translations

Can verify that `es_MX` is used if based off the language preference of the browser (see attached; note the image *is not* translated yet).

Per the log files, the img src is also acknowledging `es_MX`:

```
127.0.0.1 - - [17/May/2021 18:09:33] "GET /static/img/es_mx/how-it-works.svg HTTP/1.1" 200 -
```

For reference, the following `pybabel` commands were used:

```bash
$ pybabel extract -F ../babel.cfg . -o translations/base.pot
$ pybabel init -l es_MX en_US -i translations/base.pot -d translations
$ pybabel init -l en_US -i translations/base.pot -d translations
$ pybabel compile -d translations -f
```

And naive translations performed using [Poedit](https://poedit.net/) which seems to be commonly recommended. Note that the translations here were manually done with Google translate although it might be worth the few dollars to get a license for Poedit.

Screen Shot 2021-05-17 at 6 06 08 PM


wasade:
Note: `new_results_page.jinja2` was not extracted due to a jinja2 syntax item, and apparently pybabel is by default silent on failure. Updates coming

wasade:
....I'm not sure if we want to include this in the repo or not, but I wrote a brief script for autotranslating:

```python
import googletrans
import polib
import click


@click.command()
@click.option('--pofile', help='The file to translate',
              type=click.Path(exists=True), required=True)
@click.option('--locale', help='The locale to translate into', type=str,
              required=True)
@click.option('--output', help='A file to save into (if not specified, '
              'saves inplace)', required=False, type=click.Path(exists=False))
def trans(pofile, locale, output):
    # verify the locale name is in the path of the pofile
    assert locale in pofile
    pofile = polib.pofile(pofile)

    # example interaction based on https://github.com/Brandon1016/poTranslate (MIT license)
    translator = googletrans.Translator()
    entries = pofile.untranslated_entries()
    entries.extend(pofile.fuzzy_entries())
    for entry in pofile.untranslated_entries():
        translated = translator.translate(entry.msgid, dest=locale)
        entry.msgstr = translated.text

    if output is not None:
        pofile.save(output)
    else:
        pofile.save()


if __name__ == '__main__':
    trans()
```

Issue #139, opened 2021-05-11
Date and time on the collection page


wasade:
The sample collection page has two javascript elements allowing a user to select the date and time corresponding to their sample collection. These elements should be localized, including the validation. One point of care here is that the private API is likely assuming an en-US type representation.

PR #151, opened 2021-05-25
i18n datepicker

dhakim87:
Fixes the date portion of #139   

Localization of sample date field.  Introduces a hidden normalized field that is always mm/dd/yyyy.  Replaces validation to localize the normalized field on init, and normalize the localized field on validation.  Uses the jquery datepicker's i18n extensions for configuring the datepicker.

Far as I can tell, Spanish should be on a 24 hour clock, 8:00 AM -> 8:00, 8:00 PM -> 20:00.  Can add this in this branch as well.


dhakim87:
Asked someone from Mexico, they said AM PM is fine.

PR #153, opened 2021-05-25
en-US to en_US and using flask session

dhakim87:
en-US to en_US everywhere.  Hooked browser settings and account language settings to ApiRequests so that we query surveys in the right language.

This is blocked by the account language preferences changes.

### DIFF
--- a/microsetta_interface/server.py
+++ b/microsetta_interface/server.py
@@ -3,7 +3,7 @@ import secrets
 import logging
 
 from microsetta_interface.config_manager import SERVER_CONFIG
-from flask import jsonify, g, request
+from flask import jsonify, g
 from werkzeug.utils import redirect
 
 import connexion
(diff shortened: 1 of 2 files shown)


microsetta_interface/server.py
```diff
@@ -3,12 +3,14 @@
...
 
 from microsetta_interface.config_manager import SERVER_CONFIG
-from flask import jsonify, g, request
+from flask import jsonify, g
 from werkzeug.utils import redirect
 
 import connexion
 from flask_babel import Babel
 
+from implementation import session_locale
```
wasade, line 12:
```suggestion
from microsetta_interface.implementation import session_locale
```

wasade: ...am committing as this *should* unstick the failing test

PR #164, opened 2021-06-04
Translate Enums in Private API Model Objects

dhakim87:
Adds translations for sample.sample_site, source.source_type and survey_template.survey_template_title

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -11,7 +11,7 @@ from os import path
 from datetime import datetime
 import base64
 import functools
-import model_i18n
+from microsetta_interface import model_i18n
 
 # Authrocket uses RS256 public keys, so you can validate anywhere and safely
 # store the key in code. Obviously using this mechanism, we'd have to push code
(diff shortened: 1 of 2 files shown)


microsetta_interface/implementation.py
```diff
@@ -11,6 +11,7 @@
 from datetime import datetime
 import base64
 import functools
+import model_i18n
```
wasade, line 14:
```suggestion
from microsetta_interface import model_i18n
```

microsetta_interface/model_i18n.py
```diff
@@ -0,0 +1,70 @@
...
+    # model object types that need to be translated
+    # We define those enum values here so that babel
+    # picks them up for translation.
+
+    # Sample.sample_site
+    #     sample_site:
+    #       enum: ["Blood (skin prick)", "Saliva", "Ear wax", "Forehead",
+    #       "Fur", "Hair", "Left hand", "Left leg", "Mouth", "Nares",
+    #       "Nasal mucus", "Right hand", "Right leg", "Stool", "Tears",
+    #       "Torso", "Vaginal mucus", null]
```
wasade, line 35: would it be possible to do a `gettext` on `str(None)`, or something similar so that the null is reasonable for the user?

wasade (changes requested):

microsetta_interface/model_i18n.py
dhakim87: Replaced None with "" in the two sample fields where they show up.

wasade (approved):

PR #189, opened 2021-08-05
Uses QueryBuilder to let admins search for barcodes

dhakim87:
Jquery QueryBuilder lets admins build their own criteria to search for barcodes.  Based on their input, we build an escaped sql where clause which is sent up to the private api to perform the search.  

May be worth building the escaped sql server side so we don't run the risk of sql injection, but doesn't seem to be built into the library easily.  

DataTables runs the query and allows the end user admin to download a csv file of the barcodes.


dhakim87:
Screen Shot 2021-08-05 at 4 05 38 PM

dhakim87:
I switched to use the query-builder standalone distribution and removed the doT and extendext installs.

PR #193, opened 2021-08-20
Users can update email address

dhakim87:
Lets user update email through LoginRocket by redirecting from Account Details page.  Fixes crash on missing language when user wants to update their email.

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -57,14 +57,14 @@ ACCT_FNAME_KEY = "first_name"
 ACCT_LNAME_KEY = "last_name"
 ACCT_EMAIL_KEY = "email"
 ACCT_ADDR_KEY = "address"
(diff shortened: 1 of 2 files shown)


microsetta_interface/implementation.py
```diff
@@ -212,8 +212,6 @@ def _check_acct_prereqs(account_id, current_state=None):
         if not email_match["email_match"]:
             return NEEDS_EMAIL_CHECK, current_state
 
-        session[EMAIL_CHECK_KEY] = True
```
dhakim87, line 215: @AmandaBirmingham Thoughts on this?  session[EMAIL_CHECK_KEY] was set to True in two places, once here after every check, and once in post_update_email after the user either accepts the email change or decides not to change it (which really doesn't make a ton of sense, since it is now the email they are using for login...)  Removing this line makes us check for email changes more often, but prevents us from missing the change when it does occur.  Might indicate that this check is being done in the wrong place, possibly only needs to happen on authrocket callback, since that is where user would go after changing their email in authrocket.

PR #204, opened 2022-01-13
Setup an integration test harness

wasade:
This PR couples with biocore/microsetta-private-api#412 to provide a means for running integration tests within Github Actions (and locally as well).

There are a few pieces included here:

- a modified github actions script which includes a job to pull microsetta-private-api, install it, and execute the integration tests
- generation of JWT public/private keys
- a means to use new public/private keys at server start
- examples of fake users interacting with the system including login

The test suite provides ways to construct new users which are completely valid from the perspective of authentication, and includes the use of a de-identified user that exists in the test database to assert routing in the state machine.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,7 +90,7 @@ jobs:
         pip install -r ci/pip_requirements.txt
         pip install -e . --no-deps
         
-        git clone git@github.com:biocore/microsetta-private-api.git
+        git clone https://github.com/biocore/microsetta-private-api.git
         pushd microsetta-private-api
         pgport=${{ job.services.postgres.ports[5432] }}
         sed -i "s/self.port = 5432/self.port = $pgport/" microsetta_private_api/config_manager.py
@@ -117,6 +117,7 @@ jobs:
         popd
 
         conda activate microsetta-interface
+        source keys_for_testing.sh
         python microsetta_interface/tests/test_integration.py 2> >(tee -a stderr.log >&2)
         
         # make sure we did not skip tests, which would indicate a failure in
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,7 +114,11 @@ jobs:
         pushd microsetta-private-api
         sed -i 's/"debug": true,/"debug": true, "disable_authentication": true,/' microsetta_private_api/server_config.json
         python microsetta_private_api/server.py &
+        sleep 1
         popd
+
+        # make sure we're online
+        curl -v http://localhost:8082/api/ui
 
         conda activate microsetta-interface
         source keys_for_testing.sh
@@ -123,4 +127,4 @@ jobs:
         # make sure we did not skip tests, which would indicate a failure in
         # the harness. this test will produce an exit status of 1 if the
         # test suite reports skipped tests
-        [[ $(tail -n 1 foo | grep skipped) == "" ]];
+        [[ $(tail -n 1 stderr.log | grep skipped) == "" ]];
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,7 +90,7 @@ jobs:
         pip install -r ci/pip_requirements.txt
         pip install -e . --no-deps
         
-        git clone git@github.com/biocore/microsetta-private-api.git
+        git clone git@github.com:biocore/microsetta-private-api.git
         pushd microsetta-private-api
         pgport=${{ job.services.postgres.ports[5432] }}
         sed -i "s/self.port = 5432/self.port = $pgport/" microsetta_private_api/config_manager.py
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,7 +114,7 @@ jobs:
         pushd microsetta-private-api
         sed -i 's/"debug": true,/"debug": true, "disable_authentication": true,/' microsetta_private_api/server_config.json
         python microsetta_private_api/server.py &
-        sleep 1
+        sleep 10 # give time to start up
         popd
 
         # make sure we're online
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,7 @@ jobs:
         conda activate microsetta-interface
         conda install --yes --file ci/conda_requirements.txt
         pip install -r ci/pip_requirements.txt
+        python setup.py compile_catalog
         pip install -e . --no-deps
         
         git clone https://github.com/biocore/microsetta-private-api.git
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,10 +117,8 @@ jobs:
         sleep 10 # give time to start up
         popd
 
-        # make sure we're online
-        curl -v http://localhost:8082/api/ui
-
         conda activate microsetta-interface
+        redis-server --daemonize yes
         source keys_for_testing.sh
         python microsetta_interface/tests/test_integration.py 2> >(tee -a stderr.log >&2)
         
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,3 +46,80 @@ jobs:
       run: |
         conda activate test-microsetta-interface
         pytest
+  integration:
+    runs-on: ubuntu-latest
+      
+    # Service containers to run with `runner-job`
+    services:
+      # Label used to access the service container
+      postgres:
+        # Docker Hub image
+        image: postgres:13.4
+        env:
+          POSTGRES_DB: ag_test  
+          POSTGRES_USER: postgres  
+          POSTGRES_PASSWORD: postgres
+          
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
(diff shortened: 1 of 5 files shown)


wasade:
Actions are now setup I believe. Execution of logged in users is dependent on biocore/microsetta-private-api#412 merge

.github/workflows/main.yml
```diff
@@ -46,3 +46,84 @@ jobs:
       run: |
         conda activate test-microsetta-interface
         pytest
```
dhakim87, line 48: Oof.  This adds a lot of dependencies that have to be kept up with any changes to private api.  I guess there's no way to just say Install Private Api With Dependencies that references whatever private api's dependencies are?

wasade: isn't that done here by installing dependencies from the private API's requirements files which are part of its repo?

.github/workflows/main.yml
```diff
@@ -46,3 +46,84 @@ jobs:
...
+        
+        # establish database state
+        python microsetta_private_api/LEGACY/build_db.py
+        popd
+    - name: Run integration tests
+      shell: bash -l {0}
+      run: |
+        conda activate microsetta-private-api
+        pushd microsetta-private-api
+        sed -i 's/"debug": true,/"debug": true, "disable_authentication": true,/' microsetta_private_api/server_config.json
```
dhakim87, line 116: I'm guessing this line mysteriously breaks if we rename the "debug" key?  I'm sure there must be a command line json editor, or a super simple python script.

.github/workflows/main.yml
```diff
@@ -46,3 +46,84 @@ jobs:
...
+        python microsetta_private_api/server.py &
+        sleep 10 # give time to start up
+        popd
+
+        conda activate microsetta-interface
+        redis-server --daemonize yes
```
dhakim87, line 122: Could have sworn celery used redis for the backend, if microsetta private api requires redis, it should get started before spinning up the private api background process.

keys_for_testing.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+cwd=$(pwd)
+export MICROSETTA_INTERFACE_DEBUG_JWT_PUB=${cwd}/jwtRS256.key.pub
```
dhakim87, line 4: Since you're disabling jwt verification, I don't understand the purpose of these key files.  The jwt is made up of three separate base64 strings separated by periods.  The first one specifies the algorithm, the second is the claims, the third is the signature.  We should literally be able to base 64 some claims json and put periods on either end and it should pass with verification disabled.  

I just get worried seeing private keys anywhere.

microsetta_interface/implementation.py
```diff
@@ -39,9 +40,14 @@ class Source:
     SOURCE_TYPE_ENVIRONMENT = "environmental"
 
 
-PUB_KEY = pkg_resources.read_text(
-    'microsetta_interface',
-    "authrocket.pubkey")
+PUBKEY_ENVVAR = 'MICROSETTA_INTERFACE_DEBUG_JWT_PUB'
+if os.environ.get(PUBKEY_ENVVAR, False):
```
dhakim87, line 44: Why environment variable over the config file?

.github/workflows/main.yml
wasade: Redis is used by -interface for session management

We do not need celery for the testing here (at least at the moment) so celery workers aren't being spun up

.github/workflows/main.yml
wasade: the standard from what I've seen is `jq`, but I've never used it and we use `sed` similar to this in other areas in our gh actions so meh?

keys_for_testing.sh
wasade: -interface decodes the keys. We could also allow disabling that, instead of generating keys, but generating keys is pretty easy.

microsetta_interface/implementation.py
wasade: This is special case, and it would be nice to make it hard for production to see this stuff. But if you'd prefer it in server_config then we can do that

microsetta_interface/tests/test_integration.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def _fake_jwt(email, verified):
+    if PRIVATE_API_AVAILABLE:
+        # lie and say we're from authrocket
+        payload = {'email': email,
+                   'email_verified': verified,
+                   'iss': 'https://authrocket.com',
+                   'sub': hashlib.md5(email.encode('ascii')).hexdigest()}
+        encoded = jwt.encode(payload, PRIV_KEY, algorithm='RS256')
```
dhakim87, line 38: This seems overcomplicated.  Recommend checking out https://jwt.io/  You can build a fake token with base64, no private key file required.  Afaik, you can specify the sub as anything you like, all that matters is there's an entry for it in the database.

microsetta_interface/tests/test_integration.py
```diff
@@ -0,0 +1,161 @@
...
+USER_NEW = _fake_jwt(USER_NEW_EMAIL, True)
+
+USER_NEW_UNVERIFIED_EMAIL = 'user_unverified@foo.bar'
+USER_NEW_UNVERIFIED = _fake_jwt(USER_NEW_UNVERIFIED_EMAIL, False)
+
+USER_WITH_VALID_SAMPLE_EMAIL = "th-dq)wort@3'rey.i3l"
+USER_WITH_VALID_SAMPLE = _fake_jwt(USER_WITH_VALID_SAMPLE_EMAIL, True)
+
+
+@unittest.skipIf(not PRIVATE_API_AVAILABLE,
```
dhakim87, line 56: Do you consider not having private api available as a failure?  I feel this might silently pass if we break something in the install scripts.

microsetta_interface/tests/test_integration.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # this is a new user, so they should go to the account creation page
+        new_url = resp.headers['Location']
+        self.assertTrue(new_url.endswith('create_account'))
+
+        # make sure the creation page loads
+        resp = self.app.get(new_url)
+        self.assertEqual(resp.status_code, 200)
+        data = self._html_page(resp)
+        self.assertIn('<title>Microsetta Account Details</title>', data)
```
dhakim87, line 121: Changing the page title in private api is going to mysteriously break microsetta interface builds the next time someone tries to put in a PR, that is going to be extremely confusing for someone.

microsetta_interface/tests/test_integration.py
wasade: No. The other unit tests and assessments of the repo can be done independent of -private-api, and it would be annoying to always run -private-api locally. 

Note that the actions script in the integration test job examines the output of this run, and errors if tests were skipped

microsetta_interface/tests/test_integration.py
wasade: Titles come from -interface not -private-api. 

On the other hand, titles don't change frequently and anything we test in the content of the page is liable to change anyway?

microsetta_interface/tests/test_integration.py
wasade: Oh that's nifty, thanks, and it would be great to not generate keys and stuff

sub is used on legacy account match in -private-api so has to be something we control, and specific to the user, but open ended otherwise

dhakim87 (approved):
I think this works, but it looks hard to maintain over time.  It's breaking encapsulation of the private api module in a lot of places - it has to understand postgres and redis, it's trying to understand jwts (which should be opaque tokens to interface outside trying to grab a username or email out of them), and it's validating html pages, which aren't tightly defined by a spec.  

I've individiually commented the parts that I think are the shakiest

microsetta_interface/tests/test_integration.py
dhakim87: Ahh, as long as interface is testing its own behavior that's fine I guess.

wasade: yup yup, thanks!

microsetta_interface/tests/test_integration.py
wasade: Either there's something I completely misunderstand, or we cannot get away from using keys. Per jwt.io, the RS256 algorithm requires the use of a public and private key. Our verify in code only supports RS256. 

Generating keys is annoying but not complex, and currently works, so chalking this up to a possible change to consider in the future but not doing it here

keys_for_testing.sh
wasade: Agree re: private keys, but these are burner one-use-only, and unless a future commit changes this, not dumped to stdout anywhere

microsetta_interface/implementation.py
wasade: I'm going to keep as is, we can revise in a future refactor if needed

microsetta_interface/tests/test_integration.py
wasade: Just to note, if the install scripts are broken, we should observe a nonzero exit status on start of private-api. I'm going to close this out as well since we are testing job output to check for skips.

wasade:
Thanks Dan! And thanks for all the discussion on slack about the closing points!

PR #231, opened 2022-08-15
bristol stool chart image resizing to fit page better

raginirai553:
resized and verified page display with Cassidy
tested responsiveness

### DIFF
--- a/microsetta_interface/templates/survey.jinja2
+++ b/microsetta_interface/templates/survey.jinja2
@@ -11,8 +11,8 @@
     <link rel="stylesheet" type="text/css" href="/static/vendor/vue-form-generator-2.3.4/vfg.css">
     <style type="text/css">
         img#bristol-chart {
-            height: auto;
-            width: 50%;
+            height: 50%;
+            width: 25%;
         }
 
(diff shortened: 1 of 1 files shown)


microsetta_interface/templates/survey.jinja2
```diff
@@ -11,8 +11,8 @@
     <link rel="stylesheet" type="text/css" href="/static/vendor/vue-form-generator-2.3.4/vfg.css">
     <style type="text/css">
         img#bristol-chart {
-            height: auto;
-            width: 50%;
+            height: 50%;
```
cassidysymons, line 14: The height attribute isn't actually doing anything here, it can either be set to 'auto' or left out completely.

Also, setting width to 25% makes the image unreadable if I shrink my browser window. See attached screenshot. I think it would make more sense to set a max-width and min-width here to ensure it isn't ever excessively small or excessively large.

bristol_screenshot_20220815

cassidysymons (changes requested):

microsetta_interface/templates/survey.jinja2
raginirai553: pushed new changes, please test

PR #239, opened 2022-09-15
remove kit id from account signup

raginirai553:


microsetta_interface/tests/test_implementation.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, status_code, text, json_val, headers=None):
 
     def test_authrocket_callback_noargs_error(self):
         # We are observing exceptions on /authrocket_callback which occur with
-        # malformed requests. Currently, we are throwing a 500, but instead
+        # malformed requests. Currently,we are throwing a 500, but instead
```
cassidysymons, line 70: Can you explain why you removed the space from this comment?

microsetta_interface/implementation.py
```diff
@@ -53,6 +53,7 @@ class Source:
 ADMIN_MODE_KEY = 'admin_mode'
 LOGIN_INFO_KEY = 'login_info'
 LANG_KEY = "language"
+claim_kit_name_hint = None
```
cassidysymons, line 56: This is only used in the scope of the get_source function, please move it into that scope.

cassidysymons (changes requested):

microsetta_interface/implementation.py
raginirai553: sure doing

microsetta_interface/tests/test_implementation.py
raginirai553: no reason, it must be a typo. It is not a technically thought change

PR #241, opened 2022-09-22
Fix workflow in master-overhaul branch

cassidysymons:


cassidysymons:
@wasade Same concept as the PR you merged a few days ago, just issuing against master-overhaul rather than master this time, so overhaul PRs can see the change. Can you please review and merge if you're +1?

PR #247, opened 2022-10-12
Spain Vioscreen Bypass

cassidysymons:
For THDMI Spain, we made the decision to hide the Vioscreen option to minimize confusion, as it won't work in Spain. This behavior is working correctly on the Source page, as the Vioscreen FFQ button is not displayed for Spanish users. However, during testing, we missed the post-sample page that encourages the user to take the Vioscreen FFQ.

This change mirrors the logic of determining a Spanish user (either their account country or locale matching Spain) and, if it does, bypasses the redirect to the page suggesting they take the Vioscreen FFQ.

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -1318,7 +1318,19 @@ def post_update_sample(*, account_id=None, source_id=None, sample_id=None):
             if answer['survey_template_id'] == VIOSCREEN_ID:
                 has_ffq = True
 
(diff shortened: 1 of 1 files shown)


cassidysymons:
@wasade Can you please give this a look when you have a minute? I'd like to either rapid deploy or implement this as a hotfix when we bring the server down to update the postgres module.

wasade:
+1, good on merge and proposed deploy

PR #248, opened 2022-10-14
Adjust date validation for i18n issues

cassidysymons:
Date validation on sample details was not using the hidden normalized date field, so ambiguous dates such as October 12 (10/12 vs. 12/10 depending on locale) were running afoul of the rule about being more than 30 days in the future. Moved the validation further down the chain so it can use the normalized value and function correctly.


wasade:
+1, good catch!

PR #252, opened 2022-11-02
Changes for THDMI Japan signups

cassidysymons:
The scope of this PR is to support THDMI Japan signups using the Interested Users form on the existing platform. The actual THDMI Japan campaign will be launched on the new version of the TMI platform, so these PRs explicitly exclude Japanese translations for the consent documents and surveys.

To prevent the edge case of an existing account changing their account to Japanese and/or someone with an existing Kit ID/Activation Code from creating an account and selecting Japanese, I've prevented it from displaying as a language option in the Account Details screen.

Added:
- ja_JP locale
- workaound to prevent a user from selecting Japanese as their account language since localized consents/surveys are not available
- Japan-specific address entry panel to submit_interest.jinja2, including validation rules

### DIFF
--- a/microsetta_interface/templates/submit_interest.jinja2
+++ b/microsetta_interface/templates/submit_interest.jinja2
@@ -47,9 +47,13 @@ $(document).ready(function(){
                     required: true,
                     phoneUS: true
                 });
-            }
-
-            else if (selected_country == "ES") {
+                $("#city").rules( "add", {
+                    required: true
+                });
+                $("#state").rules( "add", {
+                    required: true
+                });
+            } else if (selected_country == "ES") {
                 for (var state of SPAIN_STATES)
                 {
                     var option = $('<option></option>')
@@ -63,7 +67,17 @@ $(document).ready(function(){
                     required: true,
                     phoneSPAIN: true
                 });
-
+                $("#city").rules( "add", {
+                    required: true
+                });
+                $("#state").rules( "add", {
+                    required: true
+                });
+            } else if (selected_country == "JP") {
+                $("#form_group_address_3").show();
+                $("#form_group_city, #form_group_state").hide();
+                $("#city").rules("remove");
+                $("#state").rules("remove");
             } else {
                 for (var country of COUNTRIES_LIST)
                 {
@@ -78,6 +92,12 @@ $(document).ready(function(){
                         $("#state").append(option);
                     }
                 }
+                $("#city").rules( "add", {
+                    required: true
+                });
+                $("#state").rules( "add", {
+                    required: true
+                });
             }
         }
     });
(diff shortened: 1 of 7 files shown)


microsetta_interface/templates/submit_interest.jinja2
```diff
@@ -63,7 +67,17 @@ $(document).ready(function(){
...
                 });
-
+                $("#city").rules( "add", {
+                    required: true
+                });
+                $("#state").rules( "add", {
+                    required: true
+                });
+            } else if (selected_country == "JP") {
+                $("#form_group_address_3").show();
```
wasade, line 77: Just to check, have example Japanese addresses been attempted in this interface? I know they differ somewhat from the structure used in the US. I suspect the third line is good enough, but I'm not positive. Related, will this third address line be exposed to Melissa for validation?

microsetta_interface/templates/submit_interest.jinja2
```diff
@@ -113,7 +133,15 @@ $(document).ready(function(){
         active_section = new_active_section;
     }
 
-    $.validator.addMethod("countryValidation", function(value, element) {
+    jQuery.validator.addMethod("over18", function(value, element) {
```
wasade, line 136: Let's punt for now, but I believe this should be campaign specific

wasade (commented):
Just a few questions

microsetta_interface/templates/submit_interest.jinja2
cassidysymons: Sagawa provided the format and I've tested that the form collects what they're looking for. I'll have Promi/Se Jin consult with them to make sure the instructions direct users to enter the right values in the right places.

As far as Melissa validation, in pre-Spain testing, we were seeing results that conflicted with what we knew to be valid or invalid addresses, so we decided to disable the address verification for the time being. I did plug Address 3 into the Melissa code before we made that decision, so the plumbing is there if/when we decided to flip it back on.

microsetta_interface/templates/submit_interest.jinja2
cassidysymons: Just opened an issue to reflect that - #253

cassidysymons: Just opened an issue to reflect that - #253

microsetta_interface/templates/submit_interest.jinja2
wasade: Okay great, thanks

microsetta_interface/templates/submit_interest.jinja2
wasade: Thanks!

PR #278, opened 2023-03-28
Pin flask-babel

cassidysymons:
@wasade And this one as well, please.

### DIFF
--- a/ci/conda_requirements.txt
+++ b/ci/conda_requirements.txt
@@ -5,5 +5,5 @@ pycryptodome
 redis-py
 redis
 pytest
-flask-babel
+flask-babel=2.0.0
 pandas


ci/conda_requirements.txt
```diff
@@ -5,5 +5,5 @@ pycryptodome
 redis-py
 redis
 pytest
-flask-babel
+flask-babel=2.0.0
```
wasade, line 8: Why that vs "flask-babel < 3.0.0" used in private api?

cassidysymons: I honestly don't recall the specific reason I had done it this way, but we install it via pip in API and conda in Interface. I seem to recall that was the root of a minor difference that caused the corresponding difference in how I'd pinned them.

wasade: let's :shipit: if passing then

PR #284, opened 2023-08-22
Overhaul Adjustments Based on Testing

cassidysymons:
Notable changes:
- Expose the address error that Melissa Data returns to the user when they're correcting their address
- Adjust re-consent process
- Allow admin accounts to bypass reconsent interface
- Fix what users can see and do without reconsenting
- Minor formatting adjustments for mobile device screen sizes
- Adjustments to results report

NOTE: Some Spanish translations are missing. They will be updated in a separate PR once Alejandra is available to review.

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -54,6 +54,7 @@ ADMIN_MODE_KEY = 'admin_mode'
 LOGIN_INFO_KEY = 'login_info'
 LANG_KEY = "language"
 RECONSENT_DECLINED_KEY = "reconsent_declined"
+SOURCE_ID = "source_id"
 
 HOME_URL = "/home"
 HELP_EMAIL = "microsetta@ucsd.edu"
@@ -523,7 +524,7 @@ def _check_acct_prereqs(account_id, current_state=None):
 def _check_source_prereqs(acct_id, source_id, current_state=None):
     SURVEY_TEMPLATE_ID_KEY = "survey_template_id"
     current_state = {} if current_state is None else current_state
-    current_state['source_id'] = source_id
+    current_state[SOURCE_ID] = source_id
 
     if not session.get(ADMIN_MODE_KEY, False):
         # Get the input source
@@ -560,7 +561,7 @@ def _check_source_prereqs(acct_id, source_id, current_state=None):
 
 def _check_biospecimen_prereqs(acct_id, source_id, current_state=None):
     current_state = {} if current_state is None else current_state
-    current_state['source_id'] = source_id
+    current_state[SOURCE_ID] = source_id
 
     if not session.get(ADMIN_MODE_KEY, False):
         # Get the input source
@@ -582,7 +583,7 @@ def _check_biospecimen_prereqs(acct_id, source_id, current_state=None):
             return NEEDS_REROUTE, current_state
 
         if consent_output["result"]:
-            session["source_id"] = source_id
+            session[SOURCE_ID] = source_id
             return NEEDS_BIOSPECIMEN_CONSENT, current_state
 
     return BIOSPECIMEN_PREREQS_MET, current_state
@@ -686,7 +687,7 @@ def prerequisite(allowed_states: list, **parameter_overrides):
             # Check relevant prereqs from those arguments
             prereqs_step, curr_state = _check_relevant_prereqs(
                 kwargs_copy.get('account_id'),
-                kwargs_copy.get('source_id')
+                kwargs_copy.get(SOURCE_ID)
             )
 
             # Route to closest sink if state doesn't match a required state
@@ -724,7 +725,7 @@ def _parse_jwt(token):
 def _route_to_closest_sink(prereqs_step, current_state):
     # print("Current Prereq Step:", prereqs_step)
     acct_id = current_state.get("account_id", None)
-    source_id = current_state.get("source_id", None)
+    source_id = current_state.get(SOURCE_ID, None)
 
     if prereqs_step == NEEDS_REROUTE:
         # where you get rerouted to depends on why you need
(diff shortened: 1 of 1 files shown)


microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -917,7 +932,7 @@
                 $("#n_archaea_background").text(result["Archaea"].toLocaleString()).addClass("microbe-count").removeClass("spinner-grow spinner-grow-sm");
             });
         retrieveNeighbors(state, 1)
-            .then(queryNeighborAndSelfMetadata(state, ["age_years", "sugary_sweets_frequency"]))
+            .then(queryNeighborAndSelfMetadata(state, ["age_cat", "sugary_sweets_frequency"]))
```
wasade, line 935: If this looks good, then let's keep it, but the corresponding variable would I think be `host_age` followign the metadata updates

wasade: ...same for below

microsetta_interface/implementation.py
```diff
@@ -1158,13 +1277,33 @@ def get_fill_source_survey(*,
     if has_error:
         return survey_output
 
+    # We need to know if their consent is current if they're trying to take an
+    # external survey. Profiles without current consent can view their surveys
+    # but we'll block them from updating them on the POST action.
+    need_reconsent = check_current_consent(account_id, source_id, "data")
```
wasade, line 1283: Should this be handled by a `prerequisite`?

wasade: ...I'm not really sure from the flow so genuine question

microsetta_interface/implementation.py
```diff
@@ -1315,6 +1464,13 @@ def post_generate_vioscreen_url(*,
                                 account_id=None,
                                 source_id=None,
                                 body=None):
+    need_reconsent = check_current_consent(account_id, source_id, "data")
```
wasade, line 1467: This does seem boilerplate so it may make sense as a prereq

microsetta_interface/implementation.py
```diff
@@ -1450,12 +1607,30 @@ def render_consent_page(account_id, source_id, form_type, sample_ids=None):
...
+        "/accounts/%s/sources/%s" %
+        (account_id, source_id)
     )
 
 
 @prerequisite([SOURCE_PREREQS_MET, BIOSPECIMEN_PREREQS_MET])
 def get_source(*, account_id=None, source_id=None):
+    # If the user is switching profiles, we need to remove the
+    # reconsent_declined variable from the session scope if it exists
+    if session.get("source_id", False) and session["source_id"] != source_id\
```
wasade, line 1630:
```suggestion
    if SOURCE_ID in session:
        if session[SOURCE_ID] != source_id:
            if session.get(RECONSENT_DECLINED_KEY, False):
```

minor but this avoids the line split. I'm not sure if the session key for source ID is stored already, if not would be nice for consistency

microsetta_interface/implementation.py
```diff
@@ -1464,14 +1639,13 @@ def get_source(*, account_id=None, source_id=None):
...
-        '/accounts/%s/source/%s/consent/%s' % (account_id, source_id, 'data'))
+    need_reconsent = check_current_consent(account_id, source_id, "data")
 
-    if has_error:
-        return consent_required
-
-    if consent_required["result"]:
-        return render_consent_page(account_id, source_id, "data")
+    if not session.get(ADMIN_MODE_KEY, False) and need_reconsent\
+            and not session.get(RECONSENT_DECLINED_KEY, False):
```
wasade, line 1645:
```suggestion
    if session.get(ADMIN_MODE_KEY, False) is False:
        if need_reconsent:
            if session.get(RECONSENT_DECLINED_KEY, False) is False:
```
similar to above but i won't die on this hill

microsetta_interface/implementation.py
```diff
@@ -3135,6 +3318,19 @@ def session_locale():
     return matched_locale
 
 
+def check_current_consent(account_id, source_id, consent_type):
```
wasade, line 3321: This likely should have `prerequisites`

wasade (changes requested):
Just a ino

microsetta_interface/templates/new_results_page.jinja2
cassidysymons: host_age doesn't work relative to the current results from microsetta-processing (just tried and reverted). We can poke at that separately.

microsetta_interface/implementation.py
cassidysymons: I didn't use prerequisites for the need_reconsent decision-making because in many of the cases, we're allowing them to access a given API path but with reduced functionality. There probably would be a way to use the prerequisite system, but I think it would end up more complex and prone to issues than the current code if/when it needs to be maintained.

microsetta_interface/implementation.py
cassidysymons: Added SOURCE_ID as a constant and adjusted throughout implementation.py, along with the adjustment to avoid line splitting.

microsetta_interface/implementation.py
cassidysymons: Adjusted.

microsetta_interface/implementation.py
cassidysymons: It's a helper function, I don't think having prerequisites would be appropriate.

microsetta_interface/templates/new_results_page.jinja2
wasade: from wgs? maybe lack of samples and okay

microsetta_interface/implementation.py
wasade: kk

microsetta_interface/implementation.py
wasade: kk

wasade (approved):

PR #286, opened 2023-08-28
Disable Japanese Locale

cassidysymons:
For the initial relaunch period, we're disabling Japanese as timing prevented us from completing a full review with the Tokyo Tech team.

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -926,7 +926,7 @@ def get_authrocket_callback(token=None, redirect_uri=None):
         # NB: Disabling Japanese for the initial relaunch period until the
         # Tokyo Tech team can review finalized translations
         if primary["language"] == "ja_JP":
-            session[LANG_KEY] = "en_US"
+            session[LANG_KEY] = EN_US_KEY
         else:
             session[LANG_KEY] = primary["language"]
     else:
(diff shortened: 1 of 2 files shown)


microsetta_interface/implementation.py
```diff
@@ -923,7 +923,12 @@ def get_authrocket_callback(token=None, redirect_uri=None):
             "account_id": primary['account_id'],
             "email": primary['email']
         }
-        session[LANG_KEY] = primary["language"]
+        # NB: Disabling Japanese for the initial relaunch period until the
+        # Tokyo Tech team can review finalized translations
+        if primary["language"] == "ja_JP":
```
wasade, line 928: Is there risk of exposure to other variations of the localization tag (e.g. just "ja")?

microsetta_interface/implementation.py
```diff
@@ -923,7 +923,12 @@ def get_authrocket_callback(token=None, redirect_uri=None):
...
+            session[LANG_KEY] = "en_US"
```
wasade, line 929:
```suggestion
            session[LANG_KEY] = EN_US_KEY
```

microsetta_interface/implementation.py
wasade, line 928:
```suggestion
        if primary["language"] == JA_JP_KEY:
```

cassidysymons: No, `primary["language"]` is coming out of the `ag.account` table which will always be the standardized locale value.

PR #294, opened 2023-10-23
Reactivate Japanese Locale

cassidysymons:
In concert with https://github.com/biocore/microsetta-private-api/pull/542:
- Reactivate Japanese locale
- Update Japanese translations of various front-end content
- Implement external reports in the UI

Notably, some of the content on the results page needed to have the `gettext()` calls adjusted to include dynamic content (the `<span>` tags into which statistics are populated). Due to sentence structure in Japanese being different from English/Spanish, there was no convenient workaround. The end result is that PO files need to be managed very carefully going forward. In a perfect world, it would be nice to find a way to keep text and variables totally separate in a way that allows for translation into all of the languages we support, but it's something we can look at when we scope out the next set of changes on the results page.


microsetta_interface/implementation.py
```diff
@@ -145,7 +145,7 @@ class Source:
 client_state = RedisCache()
 
 # Countries allowed to view the contents of the My Kits tab
-KITS_TAB_WHITELIST = {'US', 'ES', 'JP'}
+KITS_TAB_WHITELIST = {'US', 'ES', 'JP', 'MX', 'GB'}
```
wasade, line 148: no SP?

microsetta_interface/templates/account_details.jinja2
```diff
@@ -68,6 +68,32 @@
                     $("#state").empty();
                 }
 
+                // Special case to prevent users outside Japan from using Japanese as their language
```
wasade, line 71: why?

microsetta_interface/implementation.py
cassidysymons: Spain is ES, unless there's another country I'm missing?

microsetta_interface/templates/account_details.jinja2
cassidysymons: We don't have the resources to translate all future changes into Japanese in the future, so we're limiting that locale to Japanese residents in an effort to allow THDMI Japan participants to view their results, while preventing anyone in the US from using the language. We may revisit this decision in the future.

microsetta_interface/implementation.py
wasade: lol ya

microsetta_interface/templates/account_details.jinja2
wasade: if the current content works and is correct, why not defer making this special case until a point of deviation in the future?

cassidysymons: This has been removed from the PR

wasade (approved):

PR #297, opened 2023-11-14
Pin connexion[swagger-ui]

cassidysymons:
Connexion 3.0 was released earlier this month and is breaking workflow. Pinning it to < 3.0.0.


cassidysymons:
Merging in spite of the failed integration workflow, as the PR fixes the issue.

PR #299, opened 2023-11-20
Restrict Japanese to Residents of Japan

cassidysymons:


microsetta_interface/templates/account_details.jinja2
```diff
@@ -68,6 +68,32 @@
                     $("#state").empty();
                 }
 
+                // Special case to prevent users outside Japan from using Japanese as their language
+                $('#language').empty();
+                if (code == "JP") {
```
wasade, line 73: Is "code" derived from language tag rather than country? Output on here is hard via phone :/

cassidysymons: No, it's the country that the user selects for their residence.

wasade: Thanks!

wasade (approved):

PR #306, opened 2023-12-10
Allow Minors to Change age_range on Source

cassidysymons:


microsetta_interface/implementation.py
```diff
@@ -79,6 +79,11 @@ class Source:
 ACCT_WRITEABLE_KEYS = [ACCT_FNAME_KEY, ACCT_LNAME_KEY, ACCT_EMAIL_KEY,
                        ACCT_ADDR_KEY, ACCT_LANG_KEY, ACCT_TERMS_KEY]
 
+# Age groups for consent purposes, in order. Order is key as this will be
+# used to govern progression. E.g., a child may move to teen, but a teen may
+# not move to child.
+HUMAN_CONSENT_AGE_GROUPS = ["0-6", "7-12", "13-17", "18-plus"]
```
wasade, line 85:
```suggestion
HUMAN_CONSENT_AGE_GROUPS = ("0-6", "7-12", "13-17", "18-plus")
```

microsetta_interface/implementation.py
```diff
@@ -1699,6 +1703,30 @@ def decline_reconsent(*, account_id, source_id):
     )
 
 
+def get_update_age(*, account_id, source_id):
```
wasade, line 1706: No prereqs?

microsetta_interface/implementation.py
```diff
@@ -2182,6 +2235,16 @@ def get_consent_view(*, account_id=None, source_id=None, consent_type=None):
...
 
+    consent_type_parts = consent_output['consent_type'].split("_")
+    if consent_type_parts[0] == "adult":
+        age_range = "18-plus"
+    elif consent_type_parts[0] == "adolescent":
+        age_range = "13-17"
+    elif consent_type_parts[0] == "child":
+        age_range = "7-12"
+    else:
+        age_range = "0-6"
```
wasade, line 2246: not sure if this is the correct match but essentially let's be defensive and exact

```suggestion
    elif consent_type_lifestage == 'infant':
        age_range = "0-6"
    else:
        raise ValueError(f"This shouldn't happen: {consent_type_lifestage}")
```

microsetta_interface/implementation.py
```diff
@@ -2182,6 +2235,16 @@ def get_consent_view(*, account_id=None, source_id=None, consent_type=None):
         False
     )
 
+    consent_type_parts = consent_output['consent_type'].split("_")
+    if consent_type_parts[0] == "adult":
+        age_range = "18-plus"
+    elif consent_type_parts[0] == "adolescent":
+        age_range = "13-17"
+    elif consent_type_parts[0] == "child":
```
wasade, line 2243:
```suggestion
    consent_type_lifestage = consent_output['consent_type'].split("_")[0]
    if consent_type_lifestage == "adult":
        age_range = "18-plus"
    elif consent_type_lifestage == "adolescent":
        age_range = "13-17"
    elif consent_type_lifestage == "child":
```

microsetta_interface/implementation.py
```diff
@@ -3581,6 +3649,10 @@ def check_current_consent(account_id, source_id, consent_type):
     return consent_required["result"]
 
 
+def check_show_update_age(age_range):
+    return age_range != "18-plus"
```
wasade, line 3653: We really should look at having these as string literals

microsetta_interface/implementation.py
```diff
@@ -1213,7 +1218,9 @@ def get_consent_page(*, account_id=None):
         participant_name=None,
         age_range=None,
         source_id=None,
-        reconsent=reconsent
+        reconsent=reconsent,
+        cur_age_index=-1,
```
wasade, line 1222: We should avoid magic numbers here and in the templates, see comment below

microsetta_interface/templates/new_participant.jinja2
```diff
@@ -243,18 +258,24 @@
...
-        </div>
-        <div class="col-md">
-            <input type="radio" id="age-range-2" name="age-range" value="7-12" onchange="change_consent_div(this)">
-            <label for="age-range-2" class="consent-age-label">{{ tl['PARTICIPANT_FORM']['AGE_7_12'] |e}}</label>
-        </div>
-        <div class="col-md">
-            <input type="radio" id="age-range-3" name="age-range" value="13-17" onchange="change_consent_div(this)">
-            <label for="age-range-3" class="consent-age-label">{{ tl['PARTICIPANT_FORM']['AGE_13_17'] |e}}</label>
-        </div>
+            {% if cur_age_index <= 0 %}
```
wasade, line 261: This and below is testing against a magic number. Can this be changed to test against an exact value, eg `if cur_age in ('UNKNOWN', '0-6')`? Though ideally these aren't strings but instead string literals though that may be weird crossing the boundary into jinja

wasade (changes requested):

microsetta_interface/templates/new_participant.jinja2
cassidysymons: Changed the magic number situation. The age ranges are now string literals in implementation.py, but I agree that copying them into the jinja2 template would be both weird and clunky. Other than that, this PR is ready for re-review, but it won't pass integration tests until the corresponding Private API PR has been merged.

wasade:
Thanks!

cassidysymons:
@wasade Workflow is passing now after merging the corresponding Private API PR. Any further feedback or is this also good to merge?

wasade:
Nah, +1, thanks!

PR #307, opened 2023-12-11
Added admin FFQ code lookup

colts661:


microsetta_interface/templates/admin_ffq_codes.jinja2
```diff
@@ -13,13 +13,82 @@
...
+                    <i>{{ _('FFQ Registration Code') }}</i>
+                </div>
+                <div class="col-sm">
+                    <i>{{ _('Registration Code Used') }}</i>
+                </div>
+                <div class="col-sm">
+                    <i>{{ _('Registration Code Used Time') }}</i>
+                </div>
+            </div>
+            {% for code_obj in ffq_diagnostics %}
```
cassidysymons, line 52: On this output, would it be possible to ensure the columns  stay in alignment for all records? Some of the the test emails I use push the extremes, but it would be good to make sure the output gracefully handles as much as it possibly can.

ffq_lookup

cassidysymons (changes requested):

microsetta_interface/templates/admin_ffq_codes.jinja2
colts661: Pushed another commit that solves this issue. 

Note: I've also found similar overflow behavior in `admn_interested_users.jinja2` so I applied the change to that email column as well. I'm not sure if it's needed on the 2 name columns, so only 1 column was adjusted in this file.

cassidysymons (approved):

PR #310, opened 2024-01-26
Add AuthRocket Status to account search

colts661:
Only this repo needs to be updated because all account-related information is already outputted by the private API.

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -2782,8 +2782,15 @@ def get_interactive_account_search(email_query):
     if do_return:
         return email_diagnostics
 
-    accounts = [{"email": acct['email'], "account_id": acct['id']}
-                for acct in email_diagnostics['accounts']]
+    accounts = [{
+        "email": acct['email'],
+        "account_id": acct['id'],
+        "authrocket_status": (
+            "Missing"
+            if acct['auth_issuer'] is None and acct['auth_sub'] is None
+            else "Authenticated"
+        )
+    } for acct in email_diagnostics['accounts']]
     return _render_with_defaults('admin_home.jinja2',
                                  accounts=accounts)
 
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -2784,11 +2784,12 @@ def get_interactive_account_search(email_query):
 
     accounts = []
     for acct in email_diagnostics['accounts']:
-        authrocket_status = (
-            "Missing"
-            if acct['auth_issuer'] is None and acct['auth_sub'] is None
-            else "Authenticated"
-        )
+        if acct['auth_issuer'] is None and acct['auth_sub'] is None:
+            authrocket_status = "Missing"
+        elif acct['auth_issuer'] is None or acct['auth_sub'] is None:
+            authrocket_status = "Faulty - Contact Admin"
+        else:
+            authrocket_status = "Authenticated"
         acct_diag = {
             "email": acct['email'],
             "account_id": acct['id'],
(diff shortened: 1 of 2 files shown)


microsetta_interface/implementation.py
```diff
@@ -2782,8 +2782,15 @@ def get_interactive_account_search(email_query):
...
-                for acct in email_diagnostics['accounts']]
+    accounts = [{
+        "email": acct['email'],
+        "account_id": acct['id'],
+        "authrocket_status": (
+            "Missing"
+            if acct['auth_issuer'] is None and acct['auth_sub'] is None
+            else "Authenticated"
+        )
+    } for acct in email_diagnostics['accounts']]
```
wasade, line 2793: Could this be written as a regular loop without the ternary? Use of a comprehension here, with the ternary, is difficult to read

wasade: I dont think the implication is authenticated, but that we have iss and sub from authrocket, which I _think_ only occurs once the user has verified their email but it would be helpful to double check

wasade (changes requested):

microsetta_interface/implementation.py
colts661: > I dont think the implication is authenticated, but that we have iss and sub from authrocket, which I _think_ only occurs once the user has verified their email but it would be helpful to double check

@cassidysymons

colts661: @wasade simplified comprehension in a new commit; @cassidysymons could you help double check the authrocket logic, thanks!!!

microsetta_interface/implementation.py
```diff
@@ -2782,8 +2782,20 @@ def get_interactive_account_search(email_query):
...
 
-    accounts = [{"email": acct['email'], "account_id": acct['id']}
-                for acct in email_diagnostics['accounts']]
+    accounts = []
+    for acct in email_diagnostics['accounts']:
+        authrocket_status = (
+            "Missing"
+            if acct['auth_issuer'] is None and acct['auth_sub'] is None
+            else "Authenticated"
+        )
```
wasade, line 2791: Simple and verbose can be easier to read, especially if code has to be urgently looked at a year from now. How does this look? Note it also include another possibility that shouldn't happen as far as I know

```suggestion
        if acct['auth_issuer'] is None and acct['auth_sub'] is None
            authrocket_status = "Missing"
        elif acct['auth_issuer'] is None or acct['auth_sub'] is None:
            authrocket_status = 'Uncertain'
        else: 
            authrocket_status = "Verified"
        )
```

wasade: Thanks! Ternaries aren't inherently bad, but they are only syntactic sugar and better suited for very simple cases.

We don't have database level constraints on these fields that ensure one cannot be null while the other is. It shouldn't happen, but we lack a guarantee in the data model. It's inclusion is defensive consideration

cassidysymons: The general audience for this adjustment is whoever is using the help account to assist participants, so I think something more descriptive than "Uncertain" would be ideal for the unlikely state where only one field is null. Let's change that to "Faulty - Contact Admin" to maximize the chances that the information gets provided to someone who can look directly at the database and address the issue.

cassidysymons: It would also be good to express the constraint at the database level, I'll open an issue in private API. I think it would be something like `CONSTRAINT auth_state_coherent ((auth_issuer != null and auth_sub != null) or (auth_issuer is null and auth_sub is null))` but I also want to poke around the system and ensure there aren't any secondary effects we're not considering.

cassidysymons (approved):

cassidysymons:
Looks good and testing on staging indicates it's functioning as intended. Thanks for making the requested adjustments, @colts661

PR #311, opened 2024-01-29
added new sample statuses

ayobi:
Added last_update column to ag_kit_barcodes table to microsetta-private-api, [PR here](https://github.com/biocore/microsetta-private-api/pull/558). This will enable a more descriptive sample status on the 'My Reports' tab.

The new sample statuses are as followed.

- Not Yet Received (Sample has not be scanned yet)
- Sample Received - Report Pending (Sample has been scanned and has the necessary information, now just awaiting Report generation)
- Sample Received - Update Pending (Sample has been updated with new information but has not been scanned yet)
- Sample Received - Information Needed (Sample has been scanned but lacks necessary information)
- Sample Received - Information Still Needed (Sample has been updated with new information and has been scanned but it still lacks necessary information)

### DIFF
--- a/microsetta_interface/templates/reports.jinja2
+++ b/microsetta_interface/templates/reports.jinja2
@@ -90,9 +90,23 @@
                 <div class="barcode-col barcode-text">
                     {{ sample.sample_barcode }}
                 </div>
-                <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
-                    <span class="sample-label">{{ _('Not Received Yet') }}</span>
-                </div>
+                {% if sample.latest_scan_timestamp is none %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Not Yet Received ') }} </span>
+                    </div>
+                {% elif sample.last_update is none %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Sample Received - Information Needed ') }} </span>
+                    </div>
+                {% elif sample.latest_scan_timestamp > sample.last_update %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Sample Received - Information Still Needed ') }} </span>
+                    </div>
+                {% else %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Sample Received - Update Pending ') }} </span>
+                    </div>
+                {% endif %}
             </div>
             {% endfor %}
             {% for er in external_reports_kit %}
--- a/microsetta_interface/templates/reports.jinja2
+++ b/microsetta_interface/templates/reports.jinja2
@@ -90,17 +90,13 @@
                 <div class="barcode-col barcode-text">
                     {{ sample.sample_barcode }}
                 </div>
-                {% if sample.latest_scan_timestamp is none %}
+                {% if sample.sample_latest_scan_timestamp is none %}
                     <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
                         <span class="sample-label">{{ _('Not Yet Received ') }} </span>
                     </div>
-                {% elif sample.last_update is none %}
+                {% elif sample.sample_latest_sample_information_update is none or sample.sample_latest_scan_timestamp > sample.sample_latest_sample_information_update %}
                     <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
                         <span class="sample-label">{{ _('Sample Received - Information Needed ') }} </span>
-                    </div>
-                {% elif sample.latest_scan_timestamp > sample.last_update %}
-                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
-                        <span class="sample-label">{{ _('Sample Received - Information Still Needed ') }} </span>
                     </div>
                 {% else %}
                     <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">


microsetta_interface/templates/reports.jinja2
```diff
@@ -90,9 +90,23 @@
                 <div class="barcode-col barcode-text">
                     {{ sample.sample_barcode }}
                 </div>
-                <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
-                    <span class="sample-label">{{ _('Not Received Yet') }}</span>
-                </div>
+                {% if sample.latest_scan_timestamp is none %}
```
cassidysymons, line 93: I think there are a couple of variable names that need to be adjusted here. In [Sample.to_api()](https://github.com/ayobi/microsetta-private-api/blob/master/microsetta_private_api/model/sample.py#L58) in microsetta-private-api, you added `sample_latest_scan_timestamp` and `sample_last_update` but this block of code is expecting `latest_scan_timestamp` and `last_update`.

microsetta_interface/templates/reports.jinja2
```diff
@@ -90,9 +90,23 @@
...
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Not Yet Received ') }} </span>
+                    </div>
+                {% elif sample.last_update is none %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Sample Received - Information Needed ') }} </span>
+                    </div>
+                {% elif sample.latest_scan_timestamp > sample.last_update %}
```
cassidysymons, line 101: The differentiation of 'Information Needed' and 'Information Still Needed' isn't important from a UI perspective. Let's wrap this condition into the above (`elif sample.last_update is none or sample.latest_scan_timestamp > sample.last_update`).

cassidysymons (changes requested):

microsetta_interface/templates/reports.jinja2
```diff
@@ -90,9 +90,23 @@
...
-                <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
-                    <span class="sample-label">{{ _('Not Received Yet') }}</span>
-                </div>
+                {% if sample.latest_scan_timestamp is none %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Not Yet Received ') }} </span>
+                    </div>
+                {% elif sample.last_update is none %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Sample Received - Information Needed ') }} </span>
```
wasade, line 99: A sample which is received, and therefore has a scan date, could be perfect in its information in which case we do not need more information, right?

microsetta_interface/templates/reports.jinja2
```diff
@@ -90,9 +90,23 @@
...
+                    </div>
+                {% elif sample.latest_scan_timestamp > sample.last_update %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Sample Received - Information Still Needed ') }} </span>
+                    </div>
+                {% else %}
```
wasade, line 105: This block would render when `sample.latest_scan_timestamp <= sample.last_update`. For the user (or admin), I think the interpretation here is "in the processing queue" or something to that effect as it still needs to be scanned again.

Note, I'm not sure if _this_ is the correct branch to render in the (unlikely) event of a tie

microsetta_interface/templates/reports.jinja2
wasade: A recent timestamp alone doesn't imply more information is needed? It feels like this needs to be coupled with the sample status from scanning, though perhaps that's being done implicitly upstream

cassidysymons: The samples are already separated based on lock status, this loop is only through the samples that are not locked.

microsetta_interface/templates/reports.jinja2
cassidysymons: The samples are already separated based on lock status, this loop is only through the samples that are not locked.

microsetta_interface/templates/reports.jinja2
cassidysymons: I think this is the appropriate block in the event of a tie. Since this block of samples consists of ones that are not locked - which means they're not valid - that would imply that the wet lab admin will need to scan them again at some point, meaning an update is pending.

microsetta_interface/templates/reports.jinja2
```diff
@@ -90,9 +90,19 @@
...
                     {{ sample.sample_barcode }}
                 </div>
-                <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
-                    <span class="sample-label">{{ _('Not Received Yet') }}</span>
-                </div>
+                {% if sample.sample_latest_scan_timestamp is none %}
+                    <div class="barcode-col barcode-col-end" id="btn-view-{{ sample.sample_id }}">
+                        <span class="sample-label">{{ _('Not Yet Received ') }} </span>
+                    </div>
+                {% elif sample.sample_latest_sample_information_update is none or sample.sample_latest_scan_timestamp > sample.sample_latest_sample_information_update %}
```
cassidysymons, line 97: @ayobi I just tested this on our staging server and I think we're running into a timezone issue somewhere. Here's what I did:
1) Claim a sample under a user account and fill out the sample collection info (sample date/time don't really matter as long as there's a valid value there).
2) Log into microsetta-admin as an admin user, hit Scan Barcode, find the sample, and mark it sample-has-inconsistencies
3) Log back into the UI as the normal user.

At this point, the sample should show as 'Sample Received - Information Needed,' but it's rolling past this condition and hitting 'Sample Received - Update Pending.' When I inspected the database, the `latest_sample_information_update` value in `ag.ag_kit_barcodes` was `2024-02-06 17:16:48.179585-08` and the `scan_timestamp` value in `barcodes.barcode_scans` was `2024-02-06 09:19:35.086172-08`.

The most expedient solution is probably to switch the insert/update value for `latest_sample_information_update` from `utcnow()` to `now()`. Can you see if changing that in the API resolves the issue, and ensure it doesn't introduce any unintended behavior?

cassidysymons (changes requested):

microsetta_interface/templates/reports.jinja2
ayobi: Ah ok, I changed it to now() and is committed on the api [PR](https://github.com/biocore/microsetta-private-api/pull/558/commits/b82a7579ee385dd19f81de6f5e5c398d8db0d4d0). I tested it in microsetta-admin and behaved as expected.

PR #316, opened 2024-03-27
Japanese locale

ayobi:
Removed Japanese as an option for preferred language for those that select Japan as their country.


cassidysymons:
PR is approved. Not going to merge until immediately before the next deployment in case an emergency bugfix for unrelated issues is necessary before Japanese users have had sufficient time to save copies of their report.

PR #318, opened 2024-04-17
survey association lock removal

ayobi:
Removed post_claim_samples call to avoid survey to sample lock on the front end as it will be done on the [backend](https://github.com/biocore/microsetta-private-api/pull/570) before the sample metadata is pushed to qiita.


cassidysymons (changes requested):

microsetta_interface/tests/test_integration.py
```diff
@@ -367,63 +367,68 @@ def test_new_human_source_to_sample(self):
...
-
-        # verify we have our sample information
-        # get collection info
-        url = f'/accounts/{account_id}/sources/{source_id}/samples/{sample_id}'
-        resp = self.app.get(url)
-        self.assertEqual(resp.status_code, 200)
-        self.assertPageTitle(resp, 'My Kits')
-        data = self._html_page(resp)
-        self.assertIn(collection_note, data)
+        # commenting this section out as we
```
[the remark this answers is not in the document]
ayobi: Got it 👍

microsetta_interface/implementation.py
```diff
@@ -2768,14 +2769,6 @@ def post_claim_samples(*, account_id=None, source_id=None, body=None,
     if has_error:
         return survey_output
 
-    # TODO: this will have to get more nuanced when we add animal surveys?
```
cassidysymons, line 2771: I think there are still some pieces that can come out of this function, e.g. the survey_output section directly above (GitHub's UI won't let me contextually comment on lines that aren't being changed in the PR). Also, the comment above the function definition is no longer applicable.

cassidysymons (changes requested):

microsetta_interface/implementation.py
ayobi: Ah ok, so all we're left here now is just the reconsent if needed and claiming of a sample https://github.com/biocore/microsetta-interface/blob/fba3472ccff20692c92dbaed230fe7977e1905df/microsetta_interface/implementation.py#L2769-L2776

PR #325, opened 2024-11-21
report overhaul phase 2

ayobi:
For phase 2, we are implementing the new diversity score graph in the diversity tab.


microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1302,6 +1305,360 @@
...
+
+        svg.append("path")
+            .attr("d", arc)
+            .attr("transform", `translate(-27.5, 15) rotate(100)`)            
+            .style("fill", "#D9E9EF")
+            .style("stroke", "#D9E9EF")
+            .style("stroke-width", 2);
+
+        svg.append("image")
+            .attr("xlink:href", "/static/img/Line_38.svg")
```
cassidysymons, line 1334: Instead of importing SVGs for the lines, it makes more sense to use d3 to draw them. This applies to all of the instances of line SVGs.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1302,6 +1305,360 @@
...
+            .attr("x", -radius - 11)
+            .attr("y", -radius + 213);
+
+        svg.append("image")
+            .attr("xlink:href", "/static/img/Line_38.svg")
+            .style("transform", "rotate(75.5deg)")
+            .attr("x", 75)
+            .attr("y", -radius + 85);
+
+        if(diversityInSampleValue == 2.40){
```
cassidysymons, line 1345: Instead of hard-coding 2.40 as the lower boundary, please establish a variable that can be used throughout this code. We'll eventually be pulling the value from microsetta-public-api, but for now, we want it to be one place we update (the initial variable setting), rather than several places throughout the code. Same comment for 7.25 as the upper boundary.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1302,6 +1305,360 @@
...
+            );
+            var angleOffset = -54;
+            svg.append("image")
+                .attr("xlink:href", "/static/img/Line_40.svg")
+                .attr("x", x)
+                .attr("y", y)
+                .attr("transform", `rotate(${(tangentAngle * 180) / Math.PI + angleOffset}, ${x}, ${y})`);
+        }
+
+        if(diversityInSampleValue > 2.4 && diversityInSampleValue < 3.99){
```
cassidysymons, line 1395: I'm not clear on why there's different logic for various possible ranges of diversity scores. Can it be consolidated into one calculation?

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1302,6 +1305,360 @@
...
+        svg.append("text")
+            .attr("x", -radius - 50)
+            .attr("y", 145)
+            .style("fill", "#747678")
+            .style("font-size", "14px")
+            .style("line-height", "24px")
+            .style("font-family", "Mulish")
+            .style("font-weight", "400")
+            .style("font-style", "normal")
+            .text("Lowest in our dataset");
```
cassidysymons, line 1542: All static text needs to be exposed to the translation mechanism. (e.g. `.text("{{ _('Lowest in our dataset') }}");` )

microsetta_interface/templates/new_results_page.jinja2
ayobi: I tried to and adjust the proportions accordingly but after a while I still couldn't keep the indicator line to stay nicely within the ellipse for all values.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1306,358 +1306,243 @@
...
+
+    if (diversityInSampleValue >= 6.5 && diversityInSampleValue <= 6.99) {
+        const angle = scale(diversityInSampleValue);
+        drawIndicatorLine(angle, radius * 0.25, radius * 0.45, 5);
+    }
+
+    if (diversityInSampleValue >= 7.0 && diversityInSampleValue < highestDiversityScore) {
+        const angle = scale(diversityInSampleValue);
+        drawIndicatorLine(angle, radius * 0.45, radius * 0.25, -35);
+    }
```
cassidysymons, line 1413: Now that the line for the diversity score is being drawn via d3 rather than trying to position a nested SVG, can you please condense all of these conditional statements? We know what the angle is, so shouldn't it just be a matter of drawing a line from the desired inner radius to the desired outer radius? If that's not the case, please explain the issue that requires these conditionals.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1306,358 +1306,243 @@
...
+        .text("{{ _('Lowest in our dataset') }}");
+
+    svg.append("text")
+        .attr("x", -radius - 50)
+        .attr("y", 127.5)
+        .style("fill", "#006A96")
+        .style("font-size", "18px")
+        .style("font-weight", "700")
+        .style("line-height", "24px")
+        .text("2.40");
```
cassidysymons, line 1436: Please replace 2.40 with appropriate constant.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1306,358 +1306,243 @@
...
+        .text("{{ _('Highest in our dataset') }}");
+
+    svg.append("text")
+        .attr("x", radius - 42.5)
+        .attr("y", 127.5)
+        .style("fill", "#006A96")
+        .style("font-size", "18px")
+        .style("font-weight", "700")
+        .style("line-height", "24px")
+        .text("7.25");
```
cassidysymons, line 1467: Please replace 7.25 with appropriate constant.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1306,358 +1306,243 @@
...
+        .text("{{ _('Average in our dataset') }}");
+
+    svg.append("text")
+        .attr("x", 55)
+        .attr("y", -radius + 40)
+        .style("fill", "#006A96")
+        .style("font-size", "18px")
+        .style("font-weight", "700")
+        .style("line-height", "24px")
+        .text("5.31");
```
cassidysymons, line 1488: Please replace 5.31 with appropriate constant.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1306,358 +1306,243 @@
...
+
+    svg.append("text")
+        .attr("x", 110)
+        .attr("y", -radius + 185)
+        .style("fill", "#006A96")
+        .style("font-size", "18px")
+        .style("font-weight", "700")
+        .style("line-height", "24px")
+        .text("6.30");
```
cassidysymons, line 1497: The Hadza score varies by dataset - it's 6.30 for WGS and 5.20 for 16S. Please use the existing code that determines which dataset the sample report is referencing to determine the appropriate score. The below text, related to the Hadza, also needs to be positioned appropriately depending on which dataset is being used.

cassidysymons (changes requested):

microsetta_interface/templates/new_results_page.jinja2
ayobi: I'll send an email with explanation and screenshots.

cassidysymons:
Merging into a working branch for further development

PR #326, opened 2024-12-17
report overhaul phase 3

ayobi:
Added 'Your Five Most Abundant Microbes' boxplots and 'How to read a boxplot' modal for the composition tab.

Lacking:

- [ ] Stats of microbe e.g. Median of Escherichia
- [ ] 'Five Most Commonly Found Microbes in Our Dataset' boxplots
- [ ] Tabs: 'Entire dataset | 100 samples most similar to yours'


wasade:
Thanks! Could a rendered example be provided?

ayobi:
Sure thing @wasade, below are some screenshots, some zoomed out to show more
full screen zoomed out
normal size boxplots
zoomed out box plots
zoomed out modal

wasade:
Thanks!

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -594,6 +594,439 @@
...
+            }
+
+            const minVal = 0;
+            const maxVal = maxRange;
+            const q1 = relativeAbundance * 0.25;
+            const median = relativeAbundance * 0.5;
+            const q3 = relativeAbundance * 0.75;
+
+            return {
+                genus: d.Genus || d.genus || "Unknown",
```
cassidysymons, line 617:
```suggestion
                genus: d.Genus || d.genus || "{{ _('Unknown') }}",
```

All strings that are/could be displayed to the user need to be exposed to the I18N mechanism. Only tagging this instance, but please review all of the code for this issue.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -594,6 +594,439 @@
...
+            .attr("height", totalHeight);
+
+        const g = svg.append("g")
+            .attr("transform", `translate(0,${margin.top})`);
+
+        const textAnchor = "end";
+
+        mockData.forEach((d, i) => {
+            console.log(`Data for genus relativ ${d.relativeAbundance}:`, d);
+            console.log(`Index: ${i}`);
```
cassidysymons, line 653:
```suggestion
```

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -594,6 +594,439 @@
...
+        const margin = { top: 20, right: 0, bottom: 60, left: -20 };
+
+        const totalHeight = rows * panelHeight + (rows - 1) * spacingY + margin.top + margin.bottom;
+
+        d3.select("#boxplot-svg").selectAll("svg").remove();
+
+        const svg = d3.select("#boxplot-svg")
+            .append("svg")
+            .attr("width", totalWidth)
+            .attr("height", totalHeight);
```
cassidysymons, line 644: Drawing this all as one big SVG is antithetical to the responsive design approach that we've tried to take with the rest of the interface. It renders fine given sufficient screen real estate, but on a tablet or phone screen it fails to render properly.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -594,6 +594,439 @@
...
+
+        const g = svg.append("g")
+            .attr("transform", `translate(0,${margin.top})`);
+
+        const textAnchor = "end";
+
+        mockData.forEach((d, i) => {
```
cassidysymons, line 651: Rather than appending each boxplot to a single SVG, the code should render an SVG for each boxplot, which should be laid out on the screen using responsive flexbox CSS code. Two columns on full-size devices, one column on smaller screens. Additionally, the size of the boxplots should be scalable to work with smaller screens that don't have 600 horizontal pixels to work with.

If scaling all of the constituent elements of the boxplot would not render appropriately, I'm open to the idea of building the SVG with a fixed width of 600px and displaying a scaled version of the final SVG with instructions for users to click on the image to "zoom in" and view it at full resolution on smaller screens. But I'd prefer to explore a fully responsive approach.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -594,6 +594,439 @@
...
+
+                modalContent.append("p")
+                    .style("text-align", "left") 
+                    .text("Example: Microbe A is a dominant microbe, exceeding 50% of the abundance in most samples. Microbe B is typically much less abundant at less than 10%. In the case of microbe B, the plot will be much more readable if the scale only shows up to 15% rather than 100%. ");
+
+                drawExampleBoxplot();
+            }
+        });
+
+        function drawExampleBoxplot() {
```
cassidysymons, line 913: There's a lot of repeated code between this and the code that draws the actual boxplots. Can we create a helper function that just handles the actual drawing and can be fed information to draw either real or example boxplots?

cassidysymons (changes requested):

cassidysymons (approved):

cassidysymons:
Merging into working branch for further development

lmerchant:
Here are two versions of the box plot stats display order.  Let me know which one you prefer. I'm partial to Version 2 because related information is grouped together to compare.

Version 1 is with Min, Mean, and Max on row 1 and Lower Quartile, Median, and Upper Quartile on row 2. (So Min and Max, along with Lower Quartile and Upper Quartile set upper and lower bounds in each row)

boxplot_ver1


Version 2 is with Min, Mean and Lower Quartile on row 1 and Max, Median, and Upper Quartile on row 2. (So Min & Max are grouped, Mean and Median are grouped, etc.)

boxplot_ver2

Version 3 is with Min, Lower Quartile, and Median on row 1 and Mean, Upper Quartile, and max on row 2. (So values generally increase from left to right)

boxplot_ver3

lmerchant:
And for the "How to read a box plot", it is displayed differently. "Lower quartile" is referred to as 25%, and "Upper quartile" is referred to as 75%. And the boxplots display on the page do not show min and max outliers like the "how to read a box plot" indicates will appear.  I know lower and upper quartile match to 25% and 75%, but at quick glance, a user won't know that. So what if the box plots display says "25% Quartile" rather than "Lower Quartile"?

sample_box_plot

PR #334, opened 2025-01-13
Cheek Samples

cassidysymons:
This PR is intended to address the needs surrounding collecting skin samples from cheeks:

1) It adds Cheek as a valid sample site to the user interface.

2) It adds the necessary UI components to support barcode metadata, which is explained conceptually in the corresponding microsetta-private-api PR at https://github.com/biocore/microsetta-interface/pull/334

One note - integration tests will fail until the corresponding microsetta-private-api is merged into master. Unless there's a reason to do otherwise, I'd like to proceed with static code review, then revisit any necessary tweaks to the integration tests via supplemental code review if needed.


microsetta_interface/templates/sample.jinja2
```diff
@@ -31,10 +31,28 @@
...
+        function setSampleMetaVisibility(sample_site) {
+            let display_state = "";
+
+            // If the sample site is not Cheek, we will hide all of the divs and set the field values to ""
+            if (sample_site != "Cheek") {
+                display_state = "none";
+                $("#sample_site_last_washed_date").val("");
+                $("#sample_site_last_washed_time").val("");
+                $("#sample_site_last_washed_product").val("").change();
+            }
```
AmandaBirmingham, line 43: I'm a little surprised that `sample_site_last_washed_date_normalized` isn't also blanked out at the same time, for the potential case where the user initially sets something as a cheek sample but then changes it.  Or is that case handled at submission by the `monthDayYearCheekWash` validation?

AmandaBirmingham (commented):
I assume this may change a little because of the private-api refactor, but I have no concerns about the current contents, just a couple of questions.

microsetta_interface/templates/sample.jinja2
cassidysymons: Good catch, thanks.

AmandaBirmingham (approved):

PR #336, opened 2025-01-13
Update pip and conda requirements for python server to work in conda environ

lmerchant:
Remove redis-py and redis from coda requirements and add redis
    to pip_requirements.txt. It's not possible to install redis-server with
    pip or conda since it's not a python package. So the only install for
    redis is the client. Pip has a more current redis version than conda.
    With conda, redis needs to be < 5.0 and for pip, redis is 5.0.8. Having
    redis only in pip_requirements also follows the setup for
    microsetta-private-api where redis is in pip_requirements.txt and
    not in conda_requirements.txt.

### DIFF
--- a/ci/conda_requirements.txt
+++ b/ci/conda_requirements.txt
@@ -1,7 +1,7 @@
-flake8==5.0.4
-flask==2.2.5
-natsort==8.4.0
-pycryptodome==3.15.0
-babel==2.14.0
+flake8
+flask < 3.0.0
+natsort
+pycryptodome
+redis-py
 flask-babel==2.0.0
-pandas==1.3.5
+pandas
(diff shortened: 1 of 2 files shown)


wasade:
Note that `conda` is not specific to Python packages. `conda-forge` in particular is enriched for other pre-built binaries.

lmerchant:
Thanks for clarifying that conda-forge carries other packages like redis-server. I've mainly used pyenv and not conda before.

And my Ubuntu installation of redis-server was first in the path over anaconda, so I made conda come first in my path. Now I see the conda install of redis-server. 

The conda_requirements.txt file just needs redis-py < 5.0 for it to install and for the server to run.

I'll submit that change and hopefully the integration test will pass.

Issue #323, opened 2024-10-02
Barcode Selection Area Overflow


cassidysymons:
When a kit has a sufficient number of samples remaining unused, the area used to select barcode(s) to claim overflows, as seen in the screenshot below. This is a low-priority issue given the fact that our current kit configurations will never run into the issue. That said, it would be good to address at some point.
![barcode_overflow](https://github.com/user-attachments/assets/1b9c6511-eefd-4a19-a7c2-5b8df8fcc26f)

PR #337, opened 2025-01-14
style: add flex grid CSS to outer div of barcode buttons to enable wrap

lmerchant:
Make the div surrounding all the barcode buttons on the "My Kits" web page a flex grid. Now all the buttons wrap and are responsive.

Fixes Issue #323

### DIFF
--- a/microsetta_interface/templates/kits.jinja2
+++ b/microsetta_interface/templates/kits.jinja2
@@ -23,11 +23,6 @@
         .tooltip.bs-tooltip-top .tooltip-arrow::before {
             border-top-color: #0000ff !important;
         }
-        .new_kit_sample_container {
-            display: flex;
-            flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


microsetta_interface/templates/kits.jinja2
```diff
@@ -23,6 +23,11 @@
         .tooltip.bs-tooltip-top .tooltip-arrow::before {
             border-top-color: #0000ff !important;
         }
+        .new_kit_sample_container {
```
cassidysymons, line 26: Can you please move these changes into the [appropriate CSS class](https://github.com/biocore/microsetta-interface/blob/master/microsetta_interface/static/css/minimal_interface.css#L1210) in `static/css/minimal_interface.css`?

cassidysymons (changes requested):

PR #338, opened 2025-01-27
Update createDiversityGaugeChart to use d3.js radial functions in place of multiple linear functions

lmerchant:
This commit replaces the original createDiversityGaugeChart function that creates the diversity gauge chart. It uses d3.js radial functions and hard coded diversity values to display the chart rather than multiple d3.js linear functions using arbitrary values to match the figma gauge image. 

My code editor had replaced quotes, and I reverted that. So the commits to compare would be from 3 weeks ago 2fca954 d0324cf (Merge pull request #325 from ayobi/report_overhaul_phase2) with this one.


cassidysymons:
Before I dive into static review, there are some general adjustments that will probably impact some of the code:
1) It looks like Line_38.svg, Line_38_blue.svg, and Line_40.svg are no longer necessary. Can you please remove them from the branch?

2) Please add any new CSS classes to minimal_interface.css, rather than putting them into the Jinja template. I don't want to do a wholesale transition or cleanup right now, but anything new (or that you're modifying) should go into minimal_interface.css.

3) It looks like there's a bit of confusion about what lines serve what purpose. The orange line should correspond to the participant's diversity score, relative to the scale between the lowest and highest values. The line attached to "Average in our dataset" should be the same dark blue as the Hadza line. I think we'll want to reduce the outer radius of the orange line to match the outer radius of the pale blue arc to prevent it from intersecting text.

Screenshot 2025-01-27 at 2 06 22 PM

4) The visualization does not seem to scale for smaller screen sizes - can you please investigate and determine how to best approach that?

Screenshot 2025-01-27 at 2 14 13 PM

lmerchant:
Added CSS styles to make diversity gauge responsive
Updated CSS names and move colors to guage diversity chart function.
Updated gauge diversity function calculation of font width and height.
Added comments to gauge diversity function.
Renamed inner functions to be more descriptive.
Added inner function to adjust score and text position so they don't
obscure each other when they are close in value.
Removed color svg lines used in previous version of diversity gauge function

Tested diversity gauge when sample score close to other scores

guage_score_near_average
gauge_score_near_hadza

lmerchant:
Also passed in diversity values as an object to the diversity gauge function where the object is after the ajax call where diversityInSampleValue was hard coded as a response. Now hardcoded all the diversity scores to be passed to the diversity gauge function. I can change that if needed back to hard coding only the sample diversity score and hard coding the other scores inside the guage creation function.

cassidysymons:
A couple of issues:
1) It looks like the diversity score is hard-coded to 2.5 with the latest changes, which we do not want.
2) The latest changes also introduced a second copy of the participant's diversity score on the outer radius, which we also do not want. It appears there's some excess code to handle rendering the second copy of the participant's score - can you please make sure all of the relevant artifacts are removed?

lmerchant:
I removed the hard coded sample diversity score and removed the sample diversity duplicate score showing up on the outer gauge. I also added code to resize the image depending on the length of Hadza descriptive text in a language.

Screenshot 2025-02-04 at 1 32 23 PM
Screenshot 2025-02-04 at 1 31 00 PM

Issue #309, opened 2024-01-19
Register FFQ button greyed out after incorrect attempt


ayobi:
When I go to 'My FFQs and try to submit a incorrect Registration Code, the 'Register FFQ' button gets greyed out and you are unable to click it again despite it saying 'Please try again' below. See attached image. 

Expected behavior: 'Register FFQ' button remains clickable to allow more attempts. 

<img src="https://github.com/biocore/microsetta-interface/assets/17304717/a13e9746-4d69-4416-9a63-f386ff2f69f3" width=75% height=75%>

PR #341, opened 2025-02-14
Fix registration code form button from disabling on invalid entry

lmerchant:
This is regarding issue #309 Register FFQ button disabled after incorrect attempt

The problem was code disabling the button right away rather than waiting and disabling it after the form is validated to prevent multiple submissions

For the submitHandler in the Jquery Validation Plugin to get called, it requires a button inside the form with the type='submit'. This was missing.

In the implementation.py file, the function get_ajax_check_ffq_code validates the form entry, and it needs to return a JSON value of true or false since it is returning the value to a JavaScript function that doesn't understand python True or False. This function was changed to return a JSON dump of a True or False value.

Put the submitHandler of the validator outside the 'rules' object in order for it to be called.

Test using a return value of True in the function get_ajax_check_ffq_code in the implementation.py file to call the submit handler function.

Add comment in submit handler function to add code to call a function in the submit handler.


cassidysymons:
@lmerchant This pull request isn't showing any code changes. Can you double check your branch? Thanks.

lmerchant:
Drat, you're right. It's because I reverted the commit from the master branch and it isn't finding differences now. I'll work on fixing my commit on my pull request so it will show up.

lmerchant:
This branch didn't show new code so I'm closing it and creating a new pull request

PR #342, opened 2025-02-14
Fix registration code form button from disabling on invalid entry and…

lmerchant:
… retry: Issue #309

The problem was code disabling the button right away rather than waiting and disabling it after the form is validated to prevent multiple submissions

Add type='submit' to form submit button.
For the submitHandler in the Jquery Validation Plugin to get called, it requires a button inside the form with the type='submit'.

Add code to return JSON format of python True or False value in implementation.py In the implementation.py file, the function get_ajax_check_ffq_code validates the form entry, and it needs to return a JSON value of true or false since it is returning the value to a JavaScript function that doesn't understand python True or False. This function was changed to return JSON of a True or False value.

Put the submitHandler of the validator outside the 'rules' object in order for it to be called.

Test using a return value of True and False in the function get_ajax_check_ffq_code in the implementation.py file to call the submit handler function. Now the submit handler is called on a returned True value and the message of an invalid key is displayed on a returned False value.

Add comment in submit handler function to create a function call in the submit handler which is empty.

### DIFF
--- a/microsetta_interface/templates/nutrition.jinja2
+++ b/microsetta_interface/templates/nutrition.jinja2
@@ -44,43 +44,39 @@
             preventImplicitSubmission(form_name);
             preclude_whitespace('#ffq_code');
 
-            $("form[name='" + form_name + "']").on('submit', function() {
-                document.getElementById("ffq_code_button").disabled = true;
-            });
+            function handleSubmit() {
+                // TODO go to page since form validated
+                console.log('submitted form')
+            };
 
-            // Initialize form validation on the registration form.
-            // It has the name attribute "registration"
+            // Validate the registration form for a valid
+            // registration code using the jQuery validation plugin.
+            // If the form is not valid, the messages are displayed
             $("form[name='" + form_name + "']").validate({
                 // Specify validation rules
                 rules: {
-                    // The key name on the left side is the name attribute
-                    // of an input field. Validation rules are defined
-                    // on the right side
+                    // The key is the form input field name attribute and the validation rules
+                    // are defined in the value object
+                    // The route check_ffq_code_valid returns
+                    // either true or false
                     ffq_code: {
                         required: true,
                         remote: {
                             url: "/check_ffq_code_valid",
                         }
-                    },
-                    submitHandler: function (form) {
-                        form.submit();
                     }
+                },
+                submitHandler: function (form, event) {
+                    // prevent multiple submissions
+                    document.getElementById("ffq_code_button").disabled = false;
+                    handleSubmit(form);
                 },
                 messages: {
                     ffq_code: "{{ _('Your registration code is not in our system or has already been used. Please try again.') }}",
                 },
             });
-        });
 
-        /*
-        function updateButtonState(ffq_code_value) {
-            if(ffq_code_value != "") {
-                document.getElementById("ffq_code_button").disabled = false;
-            } else {
-                document.getElementById("ffq_code_button").disabled = true;
-            }
-        }
-        */
+        });
 
         function openCodePanel() {
             document.getElementById('add_code_container').style.display = '';
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -874,7 +874,8 @@ def get_ajax_check_ffq_code(ffq_code):
     except:  # noqa
         return_val = False
 
-    return return_val
+    # Convert to JSON for returning to JavaScript call
+    return json.dumps(return_val)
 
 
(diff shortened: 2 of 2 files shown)


microsetta_interface/templates/nutrition.jinja2
```diff
@@ -44,43 +44,39 @@
             preventImplicitSubmission(form_name);
             preclude_whitespace('#ffq_code');
 
-            $("form[name='" + form_name + "']").on('submit', function() {
-                document.getElementById("ffq_code_button").disabled = true;
-            });
+            function handleSubmit() {
+                // TODO go to page since form validated
+                console.log('submitted form')
```
cassidysymons, line 49: Please make sure to remove debug info before opening pull requests.

microsetta_interface/templates/nutrition.jinja2
```diff
@@ -44,43 +44,39 @@
             preventImplicitSubmission(form_name);
             preclude_whitespace('#ffq_code');
 
-            $("form[name='" + form_name + "']").on('submit', function() {
-                document.getElementById("ffq_code_button").disabled = true;
-            });
+            function handleSubmit() {
+                // TODO go to page since form validated
```
cassidysymons, line 48: This needs to be addressed before we can consider merging the pull request. As it stands, this PR breaks the functionality for the tool.

microsetta_interface/templates/nutrition.jinja2
```diff
@@ -44,43 +44,39 @@
...
                         }
-                    },
-                    submitHandler: function (form) {
-                        form.submit();
                     }
                 },
+                submitHandler: function (form, event) {
+                    // prevent multiple submissions
+                    document.getElementById("ffq_code_button").disabled = false;
+                    handleSubmit(form);
```
cassidysymons, line 72: Is a helper function necessary here?

cassidysymons (changes requested):

lmerchant:
I'll consolidate the code into the plugin submit handler and remove the console.log.

But there was no functionality in the form submit to begin with.

The original on submit code is strange because it doesn't do anything once the form is submitted. It just disables the button.

            let form_name = 'ffq_code_form';
            preventImplicitSubmission(form_name);
            preclude_whitespace('#ffq_code');

            $("form[name='" + form_name + "']").on('submit', function() {
                document.getElementById("ffq_code_button").disabled = true;
            });

And you're right, this on submit function is extra if there is a submit handler used by the jQuery validation plugin since the code inside of it can be placed in the plugin handler. 

The on submit function
            $("form[name='" + form_name + "']").on('submit', function() {
                document.getElementById("ffq_code_button").disabled = true;
            });

The jQuery validation submit handler
    submitHandler: function (form) {
      form.submit();
    }

So this the plugin submit handler should be 

    submitHandler: function (form) {
      document.getElementById("ffq_code_button").disabled = true;
    }

And then nothing is done except disabling the button

And yep, I should have removed the console.log function.


So the code doesn't break anything because nothing was done before my code changes were made.

cassidysymons:
> So the code doesn't break anything because nothing was done before my code changes were made.

That's inaccurate. I would encourage you to do the following:
1) Switch your local microsetta-interface branch to master.
2) Login using admin credentials, navigate to Admin Tools > FFQ Codes, and hit the "Generate" button next to "Generate Single Code."
3) Login using normal user credentials and use the FFQ code that was generated in step 2. The code will be deemed valid and the page will reload with a "Begin FFQ" button (you won't be able to actually open the FFQ, but that's irrelevant).

I just confirmed this behavior on both my laptop and our staging server, so if the behavior differs in your local environment, I'd be very curious to see it.

lmerchant:
OK, I’ll do that. I guess I wasn’t clear with what the code was doing. Thanks for pointing it out in more detail.

Lynne



Lynne Merchant

CMI Information Systems Analyst
Center for Microbiome Innovation
Jacobs School of Engineering University of California San Diego

lmerchant:
I added form.submit() to the jQuery validation plugin submit handler so that it will go ahead and call the POST action on the form. 

The issue of it breaking the page was due to an error in the SQL query in microsetta-admin where it was checking that a valid registration code was NULL, but it should have been checking it was NOT NULL. I'll make a pull request for that fix.

lmerchant:
I removed the pull request to modify check FFQ code database search. The nutrition page now allows a user to enter a correct FFQ code and if not, will not disable the button to submit one. I also tested the code with jest to test various cases with a valid or invalid FFQ code.

PR #343, opened 2025-02-19
SBI Pre-launch Adjustments

cassidysymons:
This pull request addresses two issues:
1) It better handles Null/None values being returned for cheek samples' barcode_meta fields.
2) It adds a prompt for users who have just edited a cheek sample to take the Skin Scoring App external survey. Given that we have no insight into whether a user has actually completed the app process (by submitting a selfie), we've opted to display this message whenever a cheek sample is edited and the skin scoring app survey is available to the user.

### DIFF
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -2518,14 +2518,14 @@ def get_update_sample(*, account_id=None, source_id=None, sample_id=None):
         if sslwd is None:
             sslwd = ""
         sample_output['barcode_meta']['sample_site_last_washed_date'] = sslwd
-        
+
         sslwt = sample_output['barcode_meta'].get(
             'sample_site_last_washed_time'
         )
         if sslwt is None:
             sslwt = ""
         sample_output['barcode_meta']['sample_site_last_washed_time'] = sslwt
-        
+
         sslwp = sample_output['barcode_meta'].get(
             'sample_site_last_washed_product'
         )
--- a/microsetta_interface/implementation.py
+++ b/microsetta_interface/implementation.py
@@ -2512,11 +2512,13 @@ def get_update_sample(*, account_id=None, source_id=None, sample_id=None):
 
     if sample_output['sample_site'] == "Cheek":
         # Format date and time to be JavaScript-friendly
-        if sample_output['barcode_meta']['sample_site_last_washed_date']\
-            is None:
+        if sample_output['barcode_meta'][
+            'sample_site_last_washed_date'
+        ] is None:
             sample_output['barcode_meta']['sample_site_last_washed_date'] = ""
-        if sample_output['barcode_meta']['sample_site_last_washed_time']\
-            is None:
+        if sample_output['barcode_meta'][
+            'sample_site_last_washed_time'
+        ] is None:
             sample_output['barcode_meta']['sample_site_last_washed_time'] = ""
 
     profile_has_samples = _check_if_source_has_samples(account_id, source_id)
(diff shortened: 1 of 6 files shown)


microsetta_interface/implementation.py
```diff
@@ -2486,6 +2510,17 @@ def get_update_sample(*, account_id=None, source_id=None, sample_id=None):
         sample_output['date'] = ""
         sample_output['time'] = ""
 
+    if sample_output['sample_site'] == "Cheek":
+        # Format date and time to be JavaScript-friendly
+        if sample_output['barcode_meta'][
+            'sample_site_last_washed_date'
```
wasade, line 2516: I'm unsure whether these keys are guaranteed to exist as the private API can return an empty dict? If they are not assured to exist then they could they be accessed through `get`?

https://github.com/biocore/microsetta-private-api/blob/d9653bc1059957f8afab23c4ca804810c365255b/microsetta_private_api/repo/sample_repo.py#L423

microsetta_interface/templates/kits.jinja2
```diff
@@ -203,6 +203,11 @@
             {{ _('Thank you for logging your sample information. It looks like you haven\'t updated your profile recently. Please review') }} <a href="/accounts/{{account_id}}/sources/{{source_id}}/take_survey?survey_template_id={{prompt_survey_id}}">{{ _('your survey responses') }}</a> {{ _('to ensure they\'re as current and complete as possible.') }}
         </div>
         {% endif %}
+        {% if prompt_skin_app %}
+        <div class="alert alert-primary alert-nutrition mt-4" role="alert">
```
wasade, line 207: Minor but this isn't nutrition?

microsetta_interface/templates/kits.jinja2
```diff
@@ -203,6 +203,11 @@
...
+            {{ _('You now have access to the Skin Scoring App survey. Please return to the') }} <a href="/accounts/{{account_id}}/sources/{{source_id}}/take_survey?survey_template_id={{prompt_survey_id}}">{{ _('My Profile tab') }}</a> {{ _('to complete it.') }}
```
wasade, line 208: Why prompt rather than redirect?

microsetta_interface/implementation.py
cassidysymons: All three keys _should_ exist for any cheek sample, and their absence would indicate an unexpected state. However, using `get` doesn't hurt anything and mitigates the minimal potential for an error.

microsetta_interface/templates/kits.jinja2
cassidysymons: Adjusted, and adjusted the other prompt from which this was copied.

microsetta_interface/templates/kits.jinja2
cassidysymons: Accessing the Skin Scoring App isn't a simple redirect to an external site. It requires the user to open a pop-up on our side with instructions, hit a button to be allocated credentials, then hit a button to open the external site. As such, prompting the user to access it feels cleaner from a UX perspective than redirecting them to the My Profile tab with the pop-up already opened.

cassidysymons:
@wasade Ready for re-review, thanks!

microsetta_interface/templates/kits.jinja2
wasade: Thanks. To check, does the pop up behave ok on mobile?

cassidysymons: Yes, the formatting is responsive and I've tested it using both Firefox and Chrome's developer-mode settings for a handful of different screen sizes.

wasade (approved):

PR #345, opened 2025-06-23
Adjust Report Template for Skin Samples

cassidysymons:
This PR moves skin samples from the old template to the new template.

### DIFF
--- a/microsetta_interface/templates/new_results_page.jinja2
+++ b/microsetta_interface/templates/new_results_page.jinja2
@@ -571,8 +571,6 @@
                     }
 
                     state.dataset_type.value = 'WGS';
-                    state.dataset_site.value = body_site;
-                    state.dataset_input.value = 'tmi-WGS-' + body_site;
                 } else {
                     for (var i = 0; i < data.length; i++) {
                         $.ajax({
@@ -585,9 +583,10 @@
                     }
 
                     state.dataset_type.value = '16S';
-                    state.dataset_site.value = body_site;
-                    state.dataset_input.value = 'tmi-16S-' + body_site;
                 }
+
+                state.dataset_site.value = body_site;
+                state.dataset_input.value = 'tmi-' + state.dataset_type.value + '-' + body_site;
             }
         });
     }
@@ -1291,15 +1290,9 @@
         }
 
         // Then, display the site-specific content for our current dataset
-        const site_specific_elements = document.getElementsByClassName("site_specific_content-" + state.dataset_site.value);
+        const site_specific_elements = document.getElementsByClassName("site_specific_content_" + state.dataset_site.value);
         for (let i = 0; i < site_specific_elements.length; i++) {
             site_specific_elements[i].style.display = "";
-        }
-
-        // Hide Microbiome Maps for skin
-        if (state.dataset_site.value === "skin") {
-            document.getElementById("microbial-map-tab").parentElement.style.display = "none";
-            document.getElementById("how_you_compare_microbiome_map").style.display = "none";
         }
     }
 
@@ -1478,7 +1471,7 @@
                     <li class="nav-item">
                         <a class="nav-link" id="taxonomy-tab" data-bs-toggle="tab" href="#taxonomy" role="tab" aria-controls="taxonomy" aria-selected="false">{{ _('Composition') }}</a>
                     </li>
-                    <li class="nav-item">
+                    <li class="nav-item site_specific_content site_specific_content_gut">
                         <a class="nav-link" id="microbial-map-tab" data-bs-toggle="tab" href="#microbial-map" role="tab" aria-controls="microbial-map" aria-selected="false">{{ _('Microbiome Map') }}</a>
                     </li>
                     <li class="nav-item">
@@ -1526,7 +1519,7 @@
                 </div>
                 </br>
 
-                <div class="how_you_compare_section" id="how_you_compare_microbiome_map">
+                <div class="how_you_compare_section site_specific_content site_specific_content_gut" id="how_you_compare_microbiome_map">
                     <h4>{{ _('Microbiome Map') }}</h4>
                     <p>
                         {{ _('Where are you on the <a class="fake-tab" href="#microbial-map" data-link="#microbial-map-tab">Microbiome Map</a>?') }}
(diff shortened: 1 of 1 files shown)


microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -984,6 +1000,11 @@
                                 "id": "one_liter_of_water_a_day_frequency",
                                 "operator": "equal",
                                 "value": "Regularly (3-5 times/week)"
+                            },
+                            {
```
cassidysymons, line 1004: These two comparisons were querying only on people who selected "Regularly (3-5 times/week)," which is counter-intuitive. Adding Daily, with agreement/approval from Se Jin.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1245,7 +1284,23 @@
             }
         );
 
+        // First, hide all of the site-specific content
```
cassidysymons, line 1287: This adds a framework for hiding/displaying certain content that should only be displayed for specific sample sites/datasets.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1245,7 +1284,23 @@
...
+            all_site_specific_elements[i].style.display = "none";
+        }
+
+        // Then, display the site-specific content for our current dataset
+        const site_specific_elements = document.getElementsByClassName("site_specific_content-" + state.dataset_site.value);
+        for (let i = 0; i < site_specific_elements.length; i++) {
+            site_specific_elements[i].style.display = "";
+        }
 
+        // Hide Microbiome Maps for skin
```
cassidysymons, line 1299: We don't have any pcoa plots for skin, so we need to hide the links to that tab.

microsetta_interface/implementation.py
```diff
@@ -2663,7 +2663,9 @@ def get_sample_results(*, account_id=None, source_id=None, sample_id=None):
     if has_error:
         return sample_output
 
-    if sample_output['sample_site'] == 'Stool':
+    if sample_output['sample_site'] in [
```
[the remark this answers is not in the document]
AmandaBirmingham: Oh, yeah, eugh.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -386,10 +386,14 @@
...
-            var pcoa_plots = ['tmi_WGS_gut_multipop', 'tmi_WGS_gut_lifestage'];
+        if(state.dataset_site.value === 'gut') {
+            if(state.dataset_type.value === 'WGS') {
+                var pcoa_plots = ['tmi_WGS_gut_multipop', 'tmi_WGS_gut_lifestage'];
+            } else {
+                var pcoa_plots = ['tmi_16S_gut_all_samples', 'tmi_16S_gut_multipop', 'tmi_16S_gut_lifestage', 'tmi_16S_gut_builtenv'];
+            }
         } else {
-            var pcoa_plots = ['tmi_16S_gut_all_samples', 'tmi_16S_gut_multipop', 'tmi_16S_gut_lifestage', 'tmi_16S_gut_builtenv'];
+            var pcoa_plots = [];
```
AmandaBirmingham, line 396: So non-gut datasets don't get PCoA plots?

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -529,11 +535,21 @@
...
+                        body_site = 'gut';
+                    } else if(data[i] === 'tmi-WGS-skin') {
+                        have_wgs = true;
+                        body_site = 'skin';
                     } else if(data[i] === 'tmi-16S-gut') {
                         have_16S = true;
+                        body_site = 'gut';
+                    } else if(data[i] === 'tmi-16S-skin') {
+                        have_16S = true;
+                        body_site = 'skin';
```
AmandaBirmingham, line 552: What is the source of the names in data?  I ask because they seem structured, and if they are, perhaps it would be preferable to parse the name (split on hyphen, check if 3 pieces, if so and if first one is tmi, use last one as body_site and check if second to last one is 16S to set have_16S, etc) than to have to manually add each new name to this if ...

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -569,8 +585,8 @@
                     }
 
                     state.dataset_type.value = '16S';
-                    state.dataset_site.value = 'gut';
-                    state.dataset_input.value = 'tmi-16S-gut';
+                    state.dataset_site.value = body_site;
```
AmandaBirmingham, line 588: I like that state.dataset_site.value is now being set with a variable instead of hard-coded! Since the value being stored in state.dataset_input.value can be expressed as 'tmi-' + state.dateset_type.value + '-' + body_site , it would be great to go a step farther and move the line to set state.dataset_site.value = body_site and the alternate line above to set state.dataset_input.value OUT of the if/else, because they will be the same for both branches.

microsetta_interface/templates/new_results_page.jinja2
AmandaBirmingham: +1

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1245,7 +1284,23 @@
...
+
+        // Then, display the site-specific content for our current dataset
+        const site_specific_elements = document.getElementsByClassName("site_specific_content-" + state.dataset_site.value);
+        for (let i = 0; i < site_specific_elements.length; i++) {
+            site_specific_elements[i].style.display = "";
+        }
 
+        // Hide Microbiome Maps for skin
+        if (state.dataset_site.value === "skin") {
+            document.getElementById("microbial-map-tab").parentElement.style.display = "none";
```
AmandaBirmingham, line 1301: Why can't the elements being handled in this if be assigned the site_specific_content class and hidden as part of the code directly above?

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1563,12 +1633,18 @@
                           </div>
                       </div>
                   </div>
-                  <p>
+                  <p class="site_specific_content site_specific_content-gut">
                       <strong>{{ _('How can I change my gut microbiome based on my results?') }}</strong>
                   </p>
-                  <p>
+                  <p class="site_specific_content site_specific_content-skin">
+                      <strong>{{ _('How can I change my skin microbiome based on my results?') }}</strong>
```
AmandaBirmingham, line 1640: I am worried about these several pieces of text (here and in a few places above) that are exactly the same for skin and gut *except* for the word "skin" or "gut".  I get that when the content is really different, it makes sense to have two different elements, but in cases like this it seems to just be creating a maintenance nightmare.  I would strongly favor incorporating the site as a variable in these pieces of text and then tagging the pieces as belonging to both site_specific_content-skin and site_specific_content-gut so they would be shown in both cases.

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1489,31 +1544,46 @@
                   <p>
                       <strong>{{ _('How do we calculate your diversity value?') }}</strong>
                   </p>
-                  <p>
+                  <p class="site_specific_content site_specific_content-gut">
```
AmandaBirmingham, line 1547: why switch from underscores to hyphen at the end of site_specific_content-gut?  This seems likely to cause confusion ...

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1666,7 +1742,7 @@
                               <img class="diversity-icon" src='/static/img/age.png'>
                           </div>
                           <div class="col-sm-8 diversity-text">
-                              {{ _('<span id="beta_diversity_age_value" class="info-loader font-weight-bold">...</span> of people with a microbiome like yours were the same age as you') }}
+                              {{ _('<span id="beta_diversity_age_value" class="info-loader font-weight-bold">...</span> of people with a microbiome like yours were a similar age as you') }}
```
AmandaBirmingham, line 1745: grammatically, this would need to be "a similar age *to* you"

microsetta_interface/templates/new_results_page.jinja2
```diff
@@ -1685,9 +1761,12 @@
                   <div class="column report-text">
                     <h3>{{ _('Your Inner Zoo (aka What\'s in your sample?)') }}</h3>
                     <br>
-                    <p>
+                    <p class="site_specific_content site_specific_content-gut">
                     {{ _('This is a table of all of the different microbes we found in your sample along with their proportions.') }}
                     </p>
+                    <p class="site_specific_content site_specific_content-skin">
+                    {{ _('This is a table of all of the different bacteria we found in your sample along with their proportions.') }}
```
AmandaBirmingham, line 1768: so the skin info includes only bacteria while the gut info includes microbes from other domains of life?

AmandaBirmingham (changes requested):
A grammar fix, a couple of strong suggestions (although I will accept counterarguments :), and some questions

microsetta_interface/templates/new_results_page.jinja2
cassidysymons: The source of the name is the `/sample/list/dataset/{sample_id}` endpoint of the public API. Unfortunately, simply using the third component of the name would not suffice, as there are values there that aren't a body site. I pulled down the values for a sample and copied them below:
```
[
  "tmi-WGS-allsamples",
  "tmi-16S-allsamples",
  "multipop-16S-gut",
  "lifestage-16S-gut",
  "tmi-16S-gut",
  "lifestage-WGS-gut",
  "tmi-WGS-gut",
  "multipop-WGS-gut",
  "builtenv-16S-allsamples"
]
```

Extra conditionals could be added to your suggestion (e.g. create an array of body_sites and verify that the third element of the name is in that array before using it as the site), but it's more or less moving the issue of conditionals around.

That said, I think there would be more elegant and comprehensive options if, down the road, we resolve the issue I brought up elsewhere of microsetta-processing using env-package and microsetta-interface using sample-site without a notion of which dataset that's in.

microsetta_interface/templates/new_results_page.jinja2
cassidysymons: I brought this edit to Se Jin's attention and we're going to harmonize both gut and skin to say "bacteria and archaea" rather than "microbes" for gut and "bacteria" for skin.

microsetta_interface/templates/new_results_page.jinja2
cassidysymons: Good catch, adjusted.

microsetta_interface/templates/new_results_page.jinja2
cassidysymons: Correct. For the time being, skin is the only non-gut sample type using this report template and it was decided that not showing any PCoA plots was preferable to investing the time in experimenting with various options to display. That decision might be re-evaluated down the road.

cassidysymons:
@AmandaBirmingham This is ready for re-review at your convenience, I believe I've either made all suggested changes or responded accordingly.

microsetta_interface/templates/new_results_page.jinja2
AmandaBirmingham: Ok, that makes sense to me.  I mean, I don't like it, but it makes sense :-/

AmandaBirmingham (approved):

